PR #239, opened 2026-03-10
KTL-4229 Implement /categories/projects endpoint

nikitavlaevjb:
- Added database migrations to create the `category` table and seed initial data.
- Added a GIN index on `project_index.markers` to optimize category-project queries.
- Introduced `CategoryController` and supporting DTOs to handle `/categories/projects` requests.
- Implemented query logic in `ProjectSearchRepository` to fetch projects grouped by categories.
- Added `CategoriesProjectsDbTest` to validate sorting, merging, and limits per category.


core/search/src/main/kotlin/io/klibs/core/search/controller/CategoryController.kt
```diff
@@ -0,0 +1,44 @@
...
+@RequestMapping("/categories")
+@Tag(name = "Categories", description = "Browse projects by categories")
+@Validated
+class CategoryController(
+    private val searchService: SearchService,
+) {
+
+    @Operation(summary = "Get projects grouped by categories")
+    @GetMapping("/projects")
+    fun getProjectsByCategories(
```
dkrasnoff, line 28: It is a new controller, and I expect it to be under authentication by default

core/search/src/main/kotlin/io/klibs/core/search/controller/CategoryController.kt
```diff
@@ -0,0 +1,44 @@
...
+        @RequestParam("limit", required = false, defaultValue = "7")
+        @Min(value = 1, message = "Limit must be >= 1")
+        @Max(value = 50, message = "Limit must be <= 50")
+        limit: Int,
+    ): CategoriesProjectsResponse {
+        val results = searchService.searchByCategories(limit)
```
dkrasnoff, line 34: I suggest not using the repository-level DTO here and instead adding an intermediate service layer DTO. 
The current solution can seem normal, but in most cases, it is poor architecture. In our project, for example, it can make a harder transition to JPA in the future.

core/search/src/main/kotlin/io/klibs/core/search/dto/repository/Category.kt
```diff
@@ -0,0 +1,6 @@
+package io.klibs.core.search
+
+data class Category(
```
dkrasnoff, line 3: Could you also separate different types of entities/layers into different packages, as we tried to do before?

core/search/src/main/kotlin/io/klibs/core/search/service/SearchService.kt
```diff
@@ -88,6 +88,11 @@ class SearchService(
     }
 
 
+    @Transactional(readOnly = true)
```
dkrasnoff, line 97: Why do we need `Transactional` here?

nikitavlaevjb: TBH I just did it the same way as other methods in this service.
My thoughts were that it is mostly for showing intent that it is read-only method and avoid accidental writes.
Also there are some potential read-only optimizations but I don't believe that they happen in our case.

core/search/src/main/kotlin/io/klibs/core/search/controller/CategoryController.kt
nikitavlaevjb: Ok, I can use service level DTO, but we still need to map the results of the query.
Would you suggest to keep `Category` projection and map it to `CategoryDto` on the service level as a step towards future JPA?
Could you explain what do you think repository-level `fun findCategoriesWithProjects(limit: Int): Map<Category, List<SearchProjectResult>>` should return instead?

dkrasnoff: `findCategoriesWithProjects` should stay the same
`searchByCategories` can return list CategoryWithProjectsDTO
and `getProjectsByCategories` should return `CategoriesProjectsResponse` with inner entities available only on this layer. In such a case you decouple rpository from service and service form api

dkrasnoff (commented):
Did you intentionally make a Merge commit, or was it accidental?

core/search/src/main/kotlin/io/klibs/core/search/dto/repository/Category.kt
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/controller/CategoryController.kt
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/controller/CategoryController.kt
nikitavlaevjb: fixed

dkrasnoff: OPTIONS method is also required, if I'm not mistaken

app/src/main/resources/db/migration/2026-Q1/2026-03-09_add_markers_gin_index_to_project_index.yml
```diff
@@ -0,0 +1,10 @@
+databaseChangeLog:
```
dkrasnoff: As I recall the flow, then we just copy all such changeSets from the latest file with creation of `project_index`, lets maybe then put this changeset to such file, so we will not forget to copy it later.

dkrasnoff (approved):

PR #245, opened 2026-04-06
KTL-4370: Use JetBrains cache redirector for Maven Central content fetching

nikitavlaevjb:


app/src/main/resources/application.yml
```diff
@@ -65,6 +65,7 @@ klibs:
         rate-limit-refill-period-sec: ${KLIBS_MAVEN_RATE_LIMIT_REFILL_PERIOD:5}
         index-endpoint: ${KLIBS_MAVEN_INDEX_ENDPOINT:https://repo1.maven.org/maven2}
         index-dir: ${KLIBS_MAVEN_INDEX_CACHE_DIR:${user.dir}/cache/maven-index}
+        content-endpoint: ${KLIBS_MAVEN_CONTENT_ENDPOINT:https://cache-redirector.jetbrains.com/repo1.maven.org/maven2/}
```
dkrasnoff, line 68: nit: https://cache-redirector.jetbrains.com/maven-central seems to be more expressive

dkrasnoff (approved):

app/src/main/resources/application.yml
nikitavlaevjb: I liked `repo1.maven.org/maven2/` more because that is similar to the link we used before :)

dkrasnoff: I just expected, that someday they can under the hood also use https://repo.maven.apache.org/maven2/ and switch between them

dkrasnoff: Anyway, you can just merge, it is only a nitpick

PR #247, opened 2026-04-08
Hotfix with processing index requests

dkrasnoff:


integrations/github/src/main/kotlin/io/klibs/integration/github/GitHubIntegrationKohsukeLibrary.kt
```diff
@@ -206,10 +206,8 @@ internal class GitHubIntegrationKohsukeLibrary(
         }
     }
 
-    private fun GHRepository.markdownRender(markdownContent: String, mode: MarkdownMode): String? {
-        return this.renderMarkdown(markdownContent, mode)
-            .readText()
-            .takeIf { it.isNotBlank() }
+    private fun GHRepository.markdownRender(markdownContent: String, mode: MarkdownMode): String {
+        return this.renderMarkdown(markdownContent, mode).readText()
```
dkrasnoff, line 210: We returned null for the empty README files. 
It raised an error in `io/klibs/core/readme/ReadmeContentBuilder.kt:24`

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -96,40 +98,48 @@ class PackageIndexingService(
...
-            val isIndividualArtifact = indexRequest.version != null
-            if (isIndividualArtifact) {
-                indexArtifact(indexRequest)
-            } else {
-                logger.error("Multi-version indexing requests are not supported")
-            }
-            indexingRequestRepository.deleteById(indexRequest.idNotNull)
-        } catch (e: Exception) {
-            logger.error("Unable to process the index request, marking the request as failed: $indexRequest", e)
-            indexingRequestRepository.markAsFailed(indexRequest.idNotNull, e.message)
```
dkrasnoff, line 131: Previously, this `indexingRequestRepository.markAsFailed(indexRequest.idNotNull, e.message)` logic was under a transaction that rolled back, so we didn't mark the request as actually failed.
I moved this logic out of transaction.

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -96,40 +98,48 @@ class PackageIndexingService(
...
+    @Transactional
+    protected fun processRequest(idToProcess: Long) {
+        val indexRequest =
```
dkrasnoff, line 136: Here, I use concurrent update to replace the previous behavior with ` FOR UPDATE SKIP LOCKED`.
The current approach is more maintainable from my point of view.

workflow.md
```diff
@@ -116,12 +116,13 @@ This block describes working on a release flow.
     * Open the project page
     * Open package page
 6. When tests pass, merge `release` into `master` via merge request.
-7. Wait for production deployment from `master`.
-8. Monitor production logs for suspicious exceptions.
-9. Perform manual smoke tests in production.
-10. Create and push tag for the current release format: `release-yyyy.mm.dd`.
-11. Force push `master` branch to `release`.
-12. Announce deployment completion in Slack: [#ked-websites-private](https://jetbrains.slack.com/archives/G017S6Z1P5Y) channel
+7. Disable indexing in [ConfigMap](https://console.intellij.net/config-map?clusterId=gke-europe-west1&namespace=klibs-test&config-map-name=klibs-features-config)
```
nikitavlaevjb: why not make it 3.3?

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -96,40 +98,48 @@ class PackageIndexingService(
     }
 
     /**
-     * Processes the package queue by retrieving requests for indexing and handling them.
+     * Processes the package queue by atomically claiming a request and processing it.
```
dkrasnoff: Hmm, seems the doc does not show the current behavior. I'll fix it

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
dkrasnoff: I forgot to add a test for this strange behavior. 
Will add it as well.

nikitavlaevjb: why not use transaction template instead of selfProvider?

nikitavlaevjb (approved):

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
dkrasnoff: I don't want to manage transaction manually

workflow.md
dkrasnoff: Just in case, I've made it after all checks passed

nikitavlaevjb: It just that it would be logical to enable indexing, check indexing, disable indexing.
Easier to not forget

dkrasnoff: Reread, you're right, it is better.
Fixed.

PR #248, opened 2026-04-10
Projects comparison main feature request

dkrasnoff:


app/src/main/kotlin/io/klibs/app/job/MaterializedViewUpdatingJob.kt
```diff
@@ -3,11 +3,13 @@ package io.klibs.app.job
 import io.klibs.core.search.service.SearchService
 import net.javacrumbs.shedlock.core.LockAssert
 import net.javacrumbs.shedlock.spring.annotation.SchedulerLock
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.scheduling.annotation.Scheduled
 import org.springframework.stereotype.Component
 import java.util.concurrent.TimeUnit
 
 @Component
+@ConditionalOnProperty("klibs.indexing", havingValue = "true")
```
nikitavlaevjb, line 12: why?

core/package/src/main/kotlin/io/klibs/core/pckg/service/PackageService.kt
```diff
@@ -72,6 +72,17 @@ class PackageService(
 
     fun findAllPackagesForSitemap(): List<SitemapPackageView> =
         packageRepository.findAllPackagesForSitemap()
+
+    fun getKotlinVersionsByProjectIds(projectIds: List<Int>): Map<Int, String?> {
+        if (projectIds.isEmpty()) return emptyMap()
+
+        return packageIndexRepository.findByProjectIdIn(projectIds)
+            .filter { it.projectId != null }
+            .groupBy { it.projectId!! }
```
nikitavlaevjb, line 81: nit: I'm almost sure we can avoid !! here, because we proved that it.projectId is not null.

core/project/src/main/kotlin/io/klibs/core/project/ProjectService.kt
```diff
@@ -34,8 +34,13 @@ class ProjectService(
 ) {
     @Transactional(readOnly = true)
     fun getProjectDetailsByName(ownerLogin: String, projectName: String): ProjectDetails? {
-        val projectEntity = projectRepository.findByNameAndOwnerLogin(projectName, ownerLogin) ?: return null
-        return buildProjectDetails(projectEntity)
+        return getProjectDetailsByNames(listOf(projectName to ownerLogin)).firstOrNull()
+    }
+
+    @Transactional(readOnly = true)
+    fun getProjectDetailsByNames(pairs: List<Pair<String, String>>): List<ProjectDetails> {
```
nikitavlaevjb, line 41: isn't pairs just a map?

app/src/test/kotlin/io/klibs/core/project/CompareControllerTest.kt
```diff
@@ -0,0 +1,164 @@
...
+        }.andReturn().response
+
+        assertEquals(
+            "{\"projects\":\"For comparison, you can only compare up to 10 projects at once and at least 1\"}",
+            response.contentAsString
+        )
+    }
+
+    @Test
+    fun `should return 400 when owner login is blank when some of the projects aren't exist`() {
```
nikitavlaevjb: wrong test name

app/src/main/resources/application.yml
```diff
@@ -96,7 +96,7 @@ management:
       show-components: always
       show-details: always
     prometheus:
-      enabled: true
+      access: read_only
```
nikitavlaevjb, line 99: why is it changed and why is it in this changeset?

app/src/main/resources/db/migration/2026-Q2/2026-04-09_recreate_project_index.sql
```diff
@@ -0,0 +1,64 @@
+DROP MATERIALIZED VIEW IF EXISTS project_index;
```
nikitavlaevjb: why create a new sql when you can reuse old one if you brought no changes here?

nikitavlaevjb (changes requested):

app/src/main/resources/application.yml
dkrasnoff: It is in a separate commit, and enabled is deprecated. 
So, I considered it a small refactoring.

nikitavlaevjb: Could you explain the change itself please?

app/src/main/resources/db/migration/2026-Q2/2026-04-09_recreate_project_index.sql
dkrasnoff: It is the pitfall, which, as I recall, you catch first :) 
We should recreate `project_index` whenever we recreate `package_index`, since the former depends on the latter.

app/src/test/kotlin/io/klibs/core/project/CompareControllerTest.kt
dkrasnoff: Oh, thanks for noticing, I think I've ctrl + c, ctrl + v here the part for the another test named `should return 400 when owner login is blank`.
Fixed.

core/project/src/main/kotlin/io/klibs/core/project/ProjectService.kt
dkrasnoff: There is a small difference. A list of pairs allows us to have duplicates in keys. In the map, we should use Map<String, List<String>> to handle duplicates.

core/package/src/main/kotlin/io/klibs/core/pckg/service/PackageService.kt
dkrasnoff: Hm, afaik I can not avoid `!!`, but instead I can use it safely, because I've checked nullability above.
If you have any idea, suggest me please, because I'm not very advanced in Kotlin sugar.

app/src/main/kotlin/io/klibs/app/job/MaterializedViewUpdatingJob.kt
dkrasnoff: Let’s pretend you didn’t see that.
Removed

app/src/main/resources/application.yml
dkrasnoff: In spring, they removed `enabled` and left only `access` [since spring 3.4](https://docs.spring.io/spring-boot/3.4/appendix/application-properties/index.html#application-properties.actuator.management.endpoint.prometheus.access). 
`access` contains 3 values: NONE, READ_ONLY and UNRESTRICTED (default)
READ_ONLY seems the most suitable for us, so I chose it

app/src/main/resources/db/migration/2026-Q2/2026-04-09_recreate_project_index.sql
nikitavlaevjb: True, but that was not the question :)
We can reuse `db/migration/2026-Q1/2026-03-06_recreate_project_index.sql` for instance, because recreation of the proejct index brings no changes to the sql script that does that

dkrasnoff: db/migration/2026-Q1/2026-03-06_recreate_project_index.sql ok, make sense

dkrasnoff: fixed

nikitavlaevjb (approved):

PR #249, opened 2026-04-13
KTL-4173: Create a service to discover new requested packages

zwiora:


app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
```diff
@@ -0,0 +1,157 @@
...
+            ?: throw badRequest("Artifact not found: maven-metadata.xml does not exist for $groupId:$artifactId")
+
+        if (version !in metadata.versioning.versions) {
+            throw badRequest(
+                "Version $version not found in maven-metadata.xml for $groupId:$artifactId"
+            )
+        }
+
+        val artifact = MavenArtifact(groupId, artifactId, version, ScraperType.MANUAL_REQUEST)
+        centralSonatypeSearchClient.getKotlinToolingMetadata(artifact)
```
nikitavlaevjb, line 75: Why not try to fetch only KotlinToolingMetadata and drop the maven-metadata?

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
```diff
@@ -0,0 +1,157 @@
...
+            Query.query("g:$groupId"),
+            Query.query("l:kotlin-tooling-metadata")
+        )
+        if (artifactId != null) {
+            query = BooleanQuery.and(query, Query.query("a:$artifactId"))
+        }
+        return query
+    }
+
+    private fun paginateSearch(query: Query): List<ArtifactData> {
```
nikitavlaevjb: Why not reuse io.klibs.integration.maven.scraper.impl.CentralSonatypeScraper#executeFindKmpArtifactsQuery?

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
```diff
@@ -0,0 +1,157 @@
...
+            results.addAll(response.page)
+            currentPage++
+        } while (response.page.isNotEmpty())
+
+        return results
+    }
+
+    private fun saveIndexRequests(mavenArtifacts: List<MavenArtifact>) {
+        val requests = mavenArtifacts.mapNotNull { artifact ->
+            val (g, a, v) = Triple(artifact.groupId, artifact.artifactId, artifact.version)
```
nikitavlaevjb: nit: using Triple here looks unconventional, good effort for readability though!
I'd do smth like this
```
  val requests = mavenArtifacts.mapNotNull { artifact ->                                                                                                                             
      with(artifact) {                                                                                                                                                               
          if (packageRepository.findByGroupIdAndArtifactIdAndVersion(groupId, artifactId, version) != null) {                                                                        
              logger.debug("Already indexed: $groupId:$artifactId:$version, skipping")                                                                                               
              return@mapNotNull null                                                                                                                                                 
          }                                                                                                                                                                          
          if (indexingRequestRepository.findByGroupIdAndArtifactIdAndVersion(groupId, artifactId, version) != null) {
              logger.debug("Already queued: $groupId:$artifactId:$version, skipping")                                                                                                
              return@mapNotNull null
          }                                                                                                                                                                          
                            
          toIndexRequest()
      }
  }
```

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
```diff
@@ -0,0 +1,157 @@
...
+            }
+            results.addAll(response.page)
+            currentPage++
+        } while (response.page.isNotEmpty())
+
+        return results
+    }
+
+    private fun saveIndexRequests(mavenArtifacts: List<MavenArtifact>) {
+        val requests = mavenArtifacts.mapNotNull { artifact ->
```
nikitavlaevjb: Maybe we should do this check before going into maven search apis

app/src/test/kotlin/io/klibs/app/indexing/RequestIndexingServiceTest.kt
```diff
@@ -0,0 +1,235 @@
...
+        whenever(centralSonatypeSearchClient.getMavenMetadata("com.example", "lib")).thenReturn(metadata)
+        whenever(centralSonatypeSearchClient.getKotlinToolingMetadata(any())).thenReturn(null)
+
+        val exception = assertThrows<ResponseStatusException> {
+            uut.requestIndexing("com.example", "lib", "1.0.0")
+        }
+
+        assertEquals(400, exception.statusCode.value())
+        assertEquals(
+            "Artifact com.example:lib:1.0.0 is not a Kotlin Multiplatform library (kotlin-tooling-metadata.json not found)",
```
nikitavlaevjb: Sometimes older versions might be not kmp, we still want to index kmp ones in case of `version == null` parameter. Make sure that path works

app/src/test/kotlin/io/klibs/app/indexing/RequestIndexingServiceTest.kt
```diff
@@ -0,0 +1,235 @@
...
+            exception.reason!!
+        )
+    }
+
+    @Test
+    @Sql(value = ["classpath:sql/RequestIndexingServiceTest/insert-into-package.sql"])
+    fun `should throw 400 when artifact is already indexed`() {
```
nikitavlaevjb: `when 'all' artifacts are already indexed`
Since tests are "documentation" for the code, please make sure they are named correctly

nikitavlaevjb (changes requested):

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
zwiora: Good point! I took the suggestion from the ticket description too literally 😅

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
zwiora: I think direct reuse of `executeFindKmpArtifactsQuery` would be difficult as-is (it's a suspend function emitting into a FlowCollector with different error handling). But that's true it reuses the main logic, so maybe I should extract this common part to a new function in `CentralSonatypeSearchClient` instead?

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
zwiora: That's a cool solution, thanks!

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
zwiora: I've moved it.

Quick note: we can make this check before calling maven search api only in case the `version` is not null. In other cases we still have to call api as we don't know if we have all the versions indexed.

app/src/test/kotlin/io/klibs/app/indexing/RequestIndexingServiceTest.kt
zwiora: In case of `version == null` I use `centralSonatypeSearchClient.searchWithThrottle` with `l:kotlin-tooling-metadata` option. If I understood correctly, as a result we get all the versions that are kmp even if older versions are not kmp. Currently in the tests I mock centralSonatypeSearchClient responses, so to test that I'd have to test the behaviour of actual maven search api.

app/src/test/kotlin/io/klibs/app/indexing/RequestIndexingServiceTest.kt
zwiora: Good point, I've fixed it and also added a test case for when some of the versions are already indexed but not all

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
nikitavlaevjb: Right, direct reuse is not looking good, but maybe it is worth trying to extract and reuse common logic.
I don't insist on it, please just try it and see if it works nicely :)

zwiora: I've tried that, but it's more complex that I expected. `CentralSonatypeScraper` uses `MavenSearchClient`, not `CentralSonatypeSearchClient` in this case, so to make it clean I think the best solution would be to introduce some additional utility class for this pagination logic only. I don't think increasing the complexity like that is worth it. After all, this common logic is a few lines of code only.

nikitavlaevjb: Check this one out, seems clean and simple to me, no? [patch.patch](https://github.com/user-attachments/files/26892873/patch.patch)

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
nikitavlaevjb: So now, if `version` is null, we can find older version that might already be saved and try to save it, right?

app/src/test/kotlin/io/klibs/app/indexing/RequestIndexingServiceTest.kt
nikitavlaevjb: You can run the indexing locally and test the actual maven search api results

nikitavlaevjb (changes requested):
See comments

nikitavlaevjb (changes requested):
@dkrasnoff accidentally re-requested review here

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
zwiora: We find all versions that are KMP from maven central, which might often fetch some versions that are already stored. That's why we later check if we have them in our database with `isAlreadyIndexedOrQueued`, and if that's true, we don't try to save them again.

nikitavlaevjb: True, sorry :)

app/src/test/kotlin/io/klibs/app/indexing/RequestIndexingServiceTest.kt
zwiora: I've run it and it works as expected :)

app/src/main/kotlin/io/klibs/app/indexing/RequestIndexingService.kt
zwiora: I've tried implementing it in the newest fixup, is that what you had in mind?

nikitavlaevjb: Just to make sure: did you just applied the patch with `git apply patch.patch` ? it seems so, but you're saying that you've "tried implementing it"?

nikitavlaevjb: It's ok if you keep this version or your version, I just wanted to make sure that we're on the same page when we are talking about this refactoring :)

zwiora: > Just to make sure: did you just applied the patch with git apply patch.patch?

Actually not, I didn't know this command so I just tried copy-pasting. Now I understand why it wasn't that easy 😅 I'll remember it for the future. Thanks for help!

nikitavlaevjb: We're done here or does my suggestion look weird?

zwiora: I like your suggestion 😊 I see I've previously missed your comment, so I've added it back. I think we're done now.

nikitavlaevjb (approved):

PR #251, opened 2026-04-15
KTL-4208: Added 2025 grant winners

nikitavlaevjb:


app/src/main/resources/db/migration/2026-Q2/2026-04-15_add_2025_grant_winners.yml
```diff
@@ -0,0 +1,30 @@
+databaseChangeLog:
+  - changeSet:
+      id: add-2025-grant-winner-markers
+      author: nikita.vlaev@jetbrains.com
+      comment: >
```
dkrasnoff: nit: you don't need >

app/src/main/resources/db/migration/2026-Q2/2026-04-15_add_2025_grant_winners.yml
```diff
@@ -0,0 +1,30 @@
...
+              JOIN scm_owner ON scm_repo.owner_id = scm_owner.id
+              WHERE (scm_repo.name = 'firebase-kotlin-sdk' AND scm_owner.login = 'GitLiveApp')
+                 OR (scm_repo.name = 'Mokkery' AND scm_owner.login = 'lupuuss')
+                 OR (scm_repo.name = 'kobweb' AND scm_owner.login = 'varabyte')
+              ON CONFLICT (project_id, type) DO NOTHING;
+
+  - changeSet:
+      id: add-grant-winner-2025-to-category
+      author: nikita.vlaev@jetbrains.com
+      comment: >
```
dkrasnoff: the same nit

core/project/src/main/kotlin/io/klibs/core/project/enums/MarkerType.kt
```diff
@@ -7,6 +7,7 @@ enum class MarkerType {
     FEATURED,
     GRANT_WINNER_2023,
     GRANT_WINNER_2024,
+    GRANT_WINNER_2025,
```
dkrasnoff, line 10: do we still need this enum, if we store these values in DB?

dkrasnoff (approved):

core/project/src/main/kotlin/io/klibs/core/project/enums/MarkerType.kt
nikitavlaevjb: Yes, we use it for DTO mapping

dkrasnoff: I've checked, it seems that it can be replaced with the String everywhere

nikitavlaevjb: Yes, however it is Enum according to the model: it has small finite amount of values with specific definitions.
I think it is clearer to leave it as enum

PR #255, opened 2026-04-21
Added version type field to packages

dkrasnoff:
Version type is required to differentiate between stable and non-stable versions. Also, we can understand which packages don't use semver at all.

It is first step in a chain of PRs for MCP

### DIFF
--- a/app/src/main/kotlin/io/klibs/app/job/PackageVersionTypeJob.kt
+++ b/app/src/main/kotlin/io/klibs/app/job/PackageVersionTypeJob.kt
@@ -1,0 +1,28 @@
+package io.klibs.app.job
+
+import io.klibs.core.pckg.service.PackageService
+import net.javacrumbs.shedlock.core.LockAssert
+import net.javacrumbs.shedlock.spring.annotation.SchedulerLock
+import org.slf4j.LoggerFactory
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+import java.util.concurrent.TimeUnit
+
+@Component
+class PackageVersionTypeJob(private val packageService: PackageService) {
+
+    private val logger = LoggerFactory.getLogger(PackageVersionTypeJob::class.java)
+
+    @Scheduled(initialDelay = 10, fixedDelay = Long.MAX_VALUE, timeUnit = TimeUnit.SECONDS)
+    @SchedulerLock(name = "fillPackageVersionTypeLock", lockAtLeastFor = "365d", lockAtMostFor = "365d")
+    fun fillVersionType() {
+        LockAssert.assertLocked()
+        logger.info("=== Starting one-time job to fill version_type for packages ===")
+
+        val totalUpdated = packageService.fillAllNullVersionTypes()
+
+        logger.info("=== Finished filling version_type for $totalUpdated packages ===")
+    }
+
+}
--- a/core/package/src/main/kotlin/io/klibs/core/pckg/service/PackageService.kt
+++ b/core/package/src/main/kotlin/io/klibs/core/pckg/service/PackageService.kt
@@ -4,6 +4,7 @@ import io.klibs.core.pckg.entity.PackageEntity
 import io.klibs.core.pckg.entity.PackageTargetEntity
 import io.klibs.core.pckg.dto.PackageDTO
 import io.klibs.core.pckg.entity.PackageIndexEntity
+import io.klibs.core.pckg.enums.VersionType
 import io.klibs.core.pckg.model.PackageDetails
 import io.klibs.core.pckg.model.PackageDeveloper
 import io.klibs.core.pckg.model.PackageLicense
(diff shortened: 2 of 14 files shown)


app/src/main/kotlin/io/klibs/app/job/PackageVersionTypeJob.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.stereotype.Component
+import java.util.concurrent.TimeUnit
+
+@Component
+class PackageVersionTypeJob(private val packageService: PackageService) {
+
+    private val logger = LoggerFactory.getLogger(PackageVersionTypeJob::class.java)
+
+    @Scheduled(initialDelay = 10, fixedDelay = Long.MAX_VALUE, timeUnit = TimeUnit.SECONDS)
+    @SchedulerLock(name = "fillPackageVersionTypeLock", lockAtLeastFor = "365d", lockAtMostFor = "365d")
```
nikitavlaevjb, line 18: Maybe just manually launch it for prod db and copy dumps?

gradle/libs.versions.toml
```diff
@@ -104,6 +106,8 @@ markdown = { module = "org.jetbrains:markdown", version.ref = "markdown" }
 
 xml-util = { module = "io.github.pdvrieze.xmlutil:serialization", version.ref = "xml-util" }
 
+kotlin-semver = { module = "io.github.z4kn4fein:semver", version.ref = "kotlin-semver" }
```
nikitavlaevjb, line 109: Does that belong in this commit?

core/package/src/main/kotlin/io/klibs/core/pckg/service/PackageService.kt
```diff
@@ -72,6 +80,28 @@ class PackageService(
...
+            val updated = selfProvider.getObject().fillNullVersionTypes(batchSize)
+            totalUpdated += updated
+            if (updated < batchSize) return totalUpdated
+            logger.info("Updated $totalUpdated packages' version_type so far.")
+        }
+    }
+
+    @Transactional(readOnly = false, propagation = Propagation.REQUIRES_NEW)
+    fun fillNullVersionTypes(batchSize: Int): Int {
+        val batch = packageRepository.findByVersionTypeIsNull(PageRequest.ofSize(batchSize))
```
nikitavlaevjb, line 96: Cool approach, you tested it already?

core/package/src/test/kotlin/io/klibs/core/pckg/service/PackageServiceTest.kt
```diff
@@ -0,0 +1,162 @@
...
+    @BeforeEach
+    fun setUp() {
+        uut = PackageService(
+            packageRepository,
+            packageIndexRepository,
+            selfProvider
+        )
+    }
+
+    private fun createTestPackageEntity(
```
nikitavlaevjb: nit: I think that commonly those construction functions are put to the bottom of the kotlin test file, no?

nikitavlaevjb (approved):

gradle/libs.versions.toml
dkrasnoff: Yeah, I use it here: https://github.com/JetBrains/klibs-io-issue-management/blob/45ded83b46ff5b82b289b08922177ff6c9ab1d8c/core/package/src/main/kotlin/io/klibs/core/pckg/enums/VersionType.kt#L12

app/src/main/kotlin/io/klibs/app/job/PackageVersionTypeJob.kt
dkrasnoff: It seems more reliable and easier to do it via a one-time job than manually. 
Also, if something goes wrong, we can rerun the job, I hope

core/package/src/main/kotlin/io/klibs/core/pckg/service/PackageService.kt
dkrasnoff: Thanks, I didn't know which approach was cool, but I've tested the task on a local machine with a copy of the prod DB.

core/package/src/test/kotlin/io/klibs/core/pckg/service/PackageServiceTest.kt
dkrasnoff: You are totally right here.
My bad, moved to bottom

core/package/src/main/kotlin/io/klibs/core/pckg/service/PackageService.kt
nikitavlaevjb: The 'PageRequest' iteration of query results :)
Personally I was not aware of such apis

gradle/libs.versions.toml
nikitavlaevjb: I mean adding this dependency should be in the same commit with the code, right?

dkrasnoff: You are right, I didn't notice that it is in the "update kotlin" commit.
Fixed.

PR #258, opened 2026-04-23
KTL-4155 Create GithubLfsDetector

zwiora:


core/readme/src/main/kotlin/io/klibs/core/readme/service/GithubLfsDetector.kt
```diff
@@ -0,0 +1,65 @@
...
+import okhttp3.Request
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+
+@Service
+class GithubLfsDetector(
+    private val okHttpClient: OkHttpClient
+) {
+
+    fun isLfsFile(rawUrl: String): Boolean {
```
nikitavlaevjb, line 40: Add a comment describing the logic behind this please, it might seem cryptic without issue context

core/readme/src/test/kotlin/io/klibs/core/readme/GithubLfsDetectorTest.kt
```diff
@@ -0,0 +1,249 @@
...
+            }
+        }
+
+        val result = GithubLfsDetector(okHttpClient).isLfsFile(url)
+
+        assertTrue(result)
+    }
+
+    @Test
+    fun `isLfsFile returns true when HEAD content length header is missing but GET call confirm it's LFS`() {
```
nikitavlaevjb, line 166: Is this a real scenario?

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksHtmlReadmeProcessor.kt
```diff
@@ -1,10 +1,15 @@
 package io.klibs.core.readme.impl
 
 import io.klibs.core.readme.ReadmeType
+import io.klibs.core.readme.service.GithubLfsDetector
 import org.springframework.stereotype.Service
 
 @Service
-class LinksHtmlReadmeProcessor : LinksBaseReadmeProcessor() {
+class LinksHtmlReadmeProcessor(
```
nikitavlaevjb, line 8: Can this parameter injection through constructor be avoided with Spring magic?

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksMarkdownReadmeProcessor.kt
```diff
@@ -65,4 +70,4 @@ class LinksMarkdownReadmeProcessor : LinksBaseReadmeProcessor() {
 private fun String.isImageFile(): Boolean {
     val imageRegex = Regex("\\.(jpg|jpeg|png|gif|bmp|tiff|webp|svg|heic|ico)$", RegexOption.IGNORE_CASE)
     return imageRegex.containsMatchIn(this.lowercase())
-}
+}
```
nikitavlaevjb: what is this change? is that newline removed?

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksBaseReadmeProcessor.kt
```diff
@@ -64,11 +70,19 @@ abstract class LinksBaseReadmeProcessor : ReadmeProcessor {
         }
     }
 
+    protected fun resolveLfsUrl(rawUrl: String): String {
```
nikitavlaevjb, line 55: This function is in the base class, but is used only for the markdown. is that fine?

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksBaseReadmeProcessor.kt
```diff
@@ -31,9 +35,11 @@ abstract class LinksBaseReadmeProcessor : ReadmeProcessor {
         srcUrlPrefix: String
     ): String {
         return readmeContent.replaceFirstGroupValue(hrefRelativeLinkRegex) { link ->
-            hrefUrlPrefix.replace(ORIGINAL_PATH_PARAMETER_NAME, link)
+            val rawUrl = hrefUrlPrefix.replace(ORIGINAL_PATH_PARAMETER_NAME, link)
```
nikitavlaevjb: could you explain why this set of changes with src and href happened? its hard to track

nikitavlaevjb (changes requested):

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksBaseReadmeProcessor.kt
zwiora: First of all, the problem we try to fix is that images that are stored using Git LFS are not displayed properly. So currently this problem affects only links after `src=`. That's why I call `resolveLfsUrl` in case of `src` links. Naturally, some `href` links might also lead to LFS pointers. I thought about replacing them too, but `LinksBaseReadmeProcessor#constructNewHrefUrlPrefix` adds additional `/blob/` part to the `href` links. I'm not sure, if we would have to keep it or get rid of it in case of LFS pointers - I don't have an example to test it on. Besides, `href` links to LFS files do work - they lead to LFS pointer instead of an actual file, but that might be actually expected. That's why I call `resolveLfsUrl` in case of `src` only.

Second of all, for `resolveLfsUrl` we need an url starting with `https://raw.githubusercontent.com`. That's why I first construct a raw link with `val rawUrl = srcUrlPrefix.replace(ORIGINAL_PATH_PARAMETER_NAME, link)` and only later add the `src=` prefix and call `resolveLfsUrl(rawUrl)`. That's also why I had to move adding a prefix away from `constructNewSrcUrlPrefix`. I've changed the behaviour of `constructNewHrefUrlPrefix` only because I wanted it to work analogously.

I feel like it all makes the code more messy, but I didn't know how to approach it differently. Maybe you have any suggestions? Or maybe I should also comment some parts?

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksBaseReadmeProcessor.kt
zwiora: I think it's used also in `LinksBaseReadmeProcessor#replaceRelativeLinks`:

```
"src=\"${resolveLfsUrl(rawUrl)}\""
```

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksMarkdownReadmeProcessor.kt
zwiora: > is that newline removed?

Yup, exactly. I guess I just deleted it at some point.

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksHtmlReadmeProcessor.kt
zwiora: I think we can use `@Autowired` for example, do you think that would look nicer? I took an inspiration here from similar cases in the project but maybe the context here is different.

core/readme/src/test/kotlin/io/klibs/core/readme/GithubLfsDetectorTest.kt
zwiora: According to the documentation, `Content-Length` is not required in a response to a HEAD request. So I've decided to take that into account and in case this field is missing we just assume the check passes - the link will be further evaluated with GET request

core/readme/src/main/kotlin/io/klibs/core/readme/service/GithubLfsDetector.kt
zwiora: I've added some comments, please let me know if I should clarify anything more :)

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksMarkdownReadmeProcessor.kt
nikitavlaevjb: I think it is not required here, github also suggests that

core/readme/src/main/kotlin/io/klibs/core/readme/service/GithubLfsDetector.kt
nikitavlaevjb: Those comments are kinda not saying anything.
I think what's confusing is why is there a head before get, why do you check content type and content length, and so on. maybe just link the documentation?

zwiora: I did link the documentation below, in the companion object where I set the values, but you're right it's not clear enough. I'll move it and rephrase the description.

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksHtmlReadmeProcessor.kt
nikitavlaevjb: okay, yeah, constructor injection is better here

core/readme/src/main/kotlin/io/klibs/core/readme/impl/LinksBaseReadmeProcessor.kt
nikitavlaevjb: Since now `src` and `href` are not baked into templates, we can go for smth like this:
Take a look and see if you like it more
[patch.patch](https://github.com/user-attachments/files/27144703/patch.patch)

zwiora: It looks way cleaner, I'll use that. Thanks!

nikitavlaevjb (approved):

PR #262, opened 2026-04-24
KTL-4247: Added KMP libraries expert skill.

dkrasnoff:
This skill should use klibs.io for getting information about kmp libraries.

### DIFF
--- a/skills/kmp-libraries-expert/SKILL.md
+++ b/skills/kmp-libraries-expert/SKILL.md
@@ -1,0 +1,83 @@
+---
+name: kotlin-multiplatform-libraries-expert
+description: Help Kotlin Multiplatform developers find KMP libraries and recommend latest/stable versions using klibs.io as the source of truth. Use when the user asks which KMP library to use, wants latest or stable versions, needs Gradle coordinates, wants to upgrade or audit KMP dependencies, or asks about KMP target support. Prefer klibs.io over memory or Maven snippets whenever version accuracy matters.
+---
+
+# KMP Klibs Version Guide
+
+Use `klibs.io` as the primary authority for Kotlin Multiplatform library discovery and version selection.
+
+## When To Use
+
+- User asks for the latest stable version of a KMP library
+- User wants Gradle coordinates for a KMP dependency
+- User asks which KMP library to choose for a capability (storage, networking, logging, DI, database, UI, etc.)
+- User wants to upgrade KMP dependencies or verify target support (iOS, Android, JVM, JS, Wasm, Kotlin/Native)
+- User asks for KMP library recommendations where version accuracy matters
+
+## Core Rules
+
+1. Treat `klibs.io` as the source of truth for versions and KMP support.
+2. Recommend the newest **stable** release unless the user explicitly wants prereleases (alpha, beta, rc, eap, etc.).
+3. Prefer exact coordinates from MCP or `klibs.io/package/…` pages over guesses from README snippets.
+4. If only a project page is available, use it to identify the project, then locate the package page for coordinates.
+5. If a library is not on `klibs.io`, say so — do not present an unverified version as fact.
+6. When the user asks for "latest", include the version number and release date if available.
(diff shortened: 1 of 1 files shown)


skills/kmp-libraries-expert/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+- **Search projects** — find KMP projects by keywords, platforms/targets, and Kotlin version. Returns project name, author, platforms, targets, packages (groupId, artifactId, description), and README.
+- **Get latest version** — retrieve latest and latest stable versions for a `groupId:artifactId`. Returns version, build tool, build tool version, and Kotlin version.
+
+MCP config:
+
+```json
+{
+    "mcpServers": {
+        "klibs": {
+            "url": "http://localhost:8080/mcp"
```
nikitavlaevjb: why is it localhost?

skills/kmp-libraries-expert/SKILL.md
```diff
@@ -0,0 +1,83 @@
+---
+name: kotlin-multiplatform-libraries-expert
+description: Help Kotlin Multiplatform developers find KMP libraries and recommend latest/stable versions using klibs.io as the source of truth. Use when the user asks which KMP library to use, wants latest or stable versions, needs Gradle coordinates, wants to upgrade or audit KMP dependencies, or asks about KMP target support. Prefer klibs.io over memory or Maven snippets whenever version accuracy matters.
```
nikitavlaevjb: I find description a bit vague.
Here is example how to make enumeration of usecases more strict:
```
description: "Use when the user asks to discover, compare, upgrade, audit, or add Kotlin Multiplatform (KMP) libraries; needs latest/stable Maven or Gradle coordinates; or needs verified KMP target/platform support. Query klibs.io or its MCP server before answering when version or target accuracy matters. Do not use for general Kotlin language help, KMP project setup, Gradle troubleshooting, or Android-only dependency questions unless KMP library metadata is needed."
```

skills/kmp-libraries-expert/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+Use `klibs.io` as the primary authority for Kotlin Multiplatform library discovery and version selection.
+
+## When To Use
+
+- User asks for the latest stable version of a KMP library
+- User wants Gradle coordinates for a KMP dependency
+- User asks which KMP library to choose for a capability (storage, networking, logging, DI, database, UI, etc.)
+- User wants to upgrade KMP dependencies or verify target support (iOS, Android, JVM, JS, Wasm, Kotlin/Native)
+- User asks for KMP library recommendations where version accuracy matters
+
```
nikitavlaevjb, line 17: I would also add 'when not to use'.
```
## When Not To Use

- User asks for general Kotlin, Compose Multiplatform, Android, Gradle, or build troubleshooting that does not require library discovery, versions, coordinates, or target support
- User asks how to configure a KMP project without choosing or validating third-party libraries
- User asks about Android-only or JVM-only dependencies and does not need KMP support verified
- User wants implementation guidance for a library already chosen and current dependency metadata is irrelevant
- User asks about publishing, indexing, or contributing to `klibs.io` itself rather than choosing a KMP library
```

nikitavlaevjb (changes requested):

skills/kmp-libraries-expert/SKILL.md
dkrasnoff: Oh, my god, it is artefacts after local testing, changed to the real api

skills/kmp-libraries-expert/SKILL.md
dkrasnoff: Replaced

skills/kmp-libraries-expert/SKILL.md
dkrasnoff: Make sense added

nikitavlaevjb (approved):

PR #292, opened 2026-04-29
KTL-4245: Explore insights meta

nikitavlaevjb:
Meta-issue tracking multiple insights features including KTL-4388 (number of dependents) and KTL-4361 (comparison feature).


app/src/main/kotlin/io/klibs/app/job/ProcessPackageIndexRequestJob.kt
```diff
@@ -9,6 +9,11 @@ import org.springframework.stereotype.Component
 import java.util.concurrent.TimeUnit
 
 @Component
+@ConditionalOnProperty(
```
dkrasnoff, line 12: Good catch 👍 

I'd only suggest not to use `matchIfMissing` and explicitly add this application.yaml with the default state.
Also, `process-queue-job` could be a little bit confusing, because it is not clear which queue we are processing.

core/package/src/main/kotlin/io/klibs/core/pckg/entity/PackageDependencyEntity.kt
```diff
@@ -0,0 +1,30 @@
...
+import jakarta.persistence.Table
+import java.io.Serializable
+
+@Embeddable
+data class PackageDependencyKey(
+    @Column(name = "package_id")
+    val packageId: Long,
+
+    @Column(name = "dep_group_id")
+    val depGroupId: String,
```
dkrasnoff: Do we really need to have `depGroupId`, `depArtifactId`, `depVersion,` or can it just be a many-to-many association, and we only need `packageId` and `depPackageId` ?

app/src/main/resources/db/migration/2026-Q2/2026-04-23_create_package_dependency.yml
```diff
@@ -0,0 +1,58 @@
...
+                  type: TEXT
+                  constraints:
+                    nullable: false
+        - addPrimaryKey:
+            tableName: package_dependency
+            columnNames: package_id, dep_group_id, dep_artifact_id, dep_version
+            constraintName: pk_package_dependency
+
+  - changeSet:
+      id: add reverse lookup index on package_dependency
```
dkrasnoff, line 35: Just in case, why do we need a reverse lookup, and does it make sense to add a version to this index as well?

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -426,7 +437,80 @@ class PackageIndexingService(
         }
     }
 
+    private fun extractAndSaveDependencies(
```
dkrasnoff: What do you think about creating a special service for all operations around "Number of dependents" metric?

dkrasnoff: I'd suggest adding a comment to this logic somewhere, that the current algorithm only counts the dependencies for the common module and does not count any dependencies for the target-specific variant.

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -426,7 +437,80 @@ class PackageIndexingService(
...
+            return
+        }
+
+        val dependencies = aggregateVariantDependencies(
+            gradleMetadata = moduleMetadata.gradleMetadata,
+            selfGroupId = mavenArtifact.groupId,
+            selfArtifactId = mavenArtifact.artifactId,
+        )
+
+        if (dependencies.isEmpty()) return
```
dkrasnoff: It seems that we should have such order here
```
        if (isReindex) {
            packageDependencyRepository.deleteAllByIdPackageId(packageId)
        }

        if (dependencies.isEmpty()) return

```

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -426,7 +437,80 @@ class PackageIndexingService(
...
+                )
+            )
+        }
+
+        packageDependencyRepository.saveAll(entities)
+        logger.debug("Saved {} dependencies for {}", entities.size, mavenArtifact)
+
+        val depGroupIds = dependencies.map { it.first }.toTypedArray()
+        val depArtifactIds = dependencies.map { it.second }.toTypedArray()
+        projectRepository.recomputeDependentCountsForDepCoords(depGroupIds, depArtifactIds)
```
dkrasnoff: Can it be an async operation?

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -426,7 +437,80 @@ class PackageIndexingService(
...
     }
 }
+
+/**
+ * Collects the union of dependency coordinates declared across all variants of a Gradle module,
+ * deduping by `(group, artifact, version)` and dropping self-references (a variant listing the
+ * root module itself). Dependencies without a resolvable version (neither `requires` nor `prefers`)
+ * are dropped.
+ */
+internal fun aggregateVariantDependencies(
```
dkrasnoff: nit: seems to be util method

app/src/test/kotlin/io/klibs/app/indexing/AggregateVariantDependenciesTest.kt
```diff
@@ -0,0 +1,185 @@
...
+                Triple("org.jetbrains.kotlinx", "kotlinx-coroutines-core-jvm", "1.10.1"),
+                Triple("org.jetbrains.kotlinx", "kotlinx-coroutines-core-js", "1.10.1"),
+                Triple("org.jetbrains.kotlinx", "atomicfu", "0.23.0"),
+            ),
+            result,
+        )
+    }
+
+    @Test
+    fun `self reference across variants is filtered out`() {
```
dkrasnoff: Just wondering, is it a real case to have a self-reference in the variant?

app/src/test/kotlin/io/klibs/app/indexing/AggregateVariantDependenciesTest.kt
```diff
@@ -0,0 +1,185 @@
...
+        )
+        val result = aggregateVariantDependencies(md, selfGroupId = "io.klibs", selfArtifactId = "some-lib")
+        assertEquals(
+            setOf(Triple("org.jetbrains.kotlin", "kotlin-stdlib", "2.1.0")),
+            result,
+        )
+    }
+
+    @Test
+    fun `dependency without requires or prefers is dropped`() {
```
dkrasnoff: Hm, interesting catch.
Am I understanding correctly that in variants, we cannot have an exact dependency but can only have something that says "dependency request"?

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -426,7 +437,80 @@ class PackageIndexingService(
...
+    gradleMetadata: GradleMetadata,
+    selfGroupId: String,
+    selfArtifactId: String,
+): Set<Triple<String, String, String>> {
```
dkrasnoff: nit: I'd prefer a data class instead of Triple<String, String, String> here

app/src/main/resources/db/migration/db.changelog-master.yml
```diff
@@ -151,4 +151,12 @@ databaseChangeLog:
       file: db/migration/2026-Q2/2026-04-09_package_index_add_kotlin_version.yml
   - include:
       file: db/migration/2026-Q2/2026-04-15_add_2025_grant_winners.yml
+  - include:
```
dkrasnoff, line 154: check the order of files pls, for some reason `2026-04-10` is after `2026-04-24`

app/src/main/kotlin/io/klibs/app/job/RefreshDependentCountJob.kt
```diff
@@ -0,0 +1,31 @@
...
+@Component
+class RefreshDependentCountJob(
+    private val projectRepository: ProjectRepository,
+) {
+
+    @Scheduled(initialDelay = 0, fixedRate = 6, timeUnit = TimeUnit.HOURS)
+    @SchedulerLock(name = "refreshDependentCountLock", lockAtMostFor = "1h")
+    fun refreshDependentCounts() {
+        LockAssert.assertLocked()
+        val started = System.currentTimeMillis()
```
dkrasnoff: Good idea to count the execution time.
Why don't you put it in a metric so we can add it to our dashboard?

dkrasnoff: If not for consistency it could be `Instant.now()`

core/package/src/main/kotlin/io/klibs/core/pckg/entity/PackageDependencyEntity.kt
nikitavlaevjb: We might learn about dependency before we learn about the packаge

app/src/main/resources/db/migration/2026-Q2/2026-04-23_create_package_dependency.yml
nikitavlaevjb: Because we look up amount of packages that depend on certain other package. that is a reverse lookup

core/package/src/main/kotlin/io/klibs/core/pckg/entity/PackageDependencyEntity.kt
nikitavlaevjb: TODO try createing unfinished packages and linking them here.
Keep in mind that 'number of dependents' is a link between projects

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
nikitavlaevjb: Create a service that encapsulates pom handling logic and make PackageIndexingService only orchestrator of other services

nikitavlaevjb: Also use pom instead of module file

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
nikitavlaevjb: delete this

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
nikitavlaevjb: delete this

app/src/test/kotlin/io/klibs/app/indexing/AggregateVariantDependenciesTest.kt
nikitavlaevjb: delete this

app/src/test/kotlin/io/klibs/app/indexing/AggregateVariantDependenciesTest.kt
nikitavlaevjb: delete this

app/src/main/resources/db/migration/db.changelog-master.yml
nikitavlaevjb: rename migrations

dkrasnoff (approved):

app/src/main/resources/db/migration/2026-Q2/2026-04-23_create_maven_artifact.yml
```diff
@@ -0,0 +1,195 @@
...
+              - column:
+                  name: version
+                  type: TEXT
+                  constraints:
+                    nullable: false
+        - sql:
+            sql: |
+              INSERT INTO maven_artifact (id, group_id, artifact_id, version)
+              SELECT
+                  row_number() OVER (ORDER BY group_id, artifact_id, version) AS id,
```
nikitavlaevjb, line 40: Will such backfill advance the id sequence?
Why not take actual package_id here?

app/src/main/resources/db/migration/2026-Q2/2026-04-23_create_maven_artifact.yml
```diff
@@ -0,0 +1,195 @@
...
+                AND p.maven_artifact_id IS NULL;
+
+  - changeSet:
+      id: add package.maven_artifact_id foreign key
+      author: dmitrii.krasnov@jetbrains.com
+      preConditions:
+        - onFail: MARK_RAN
+        - not:
+            foreignKeyConstraintExists:
+              foreignKeyName: fk_package_maven_artifact_id
```
nikitavlaevjb, line 154: So is 'package' table a source of truth or the 'maven_artifact'?
That tells us which way should cascade delete go

core/package/src/main/kotlin/io/klibs/core/pckg/service/MavenArtifactService.kt
```diff
@@ -0,0 +1,84 @@
...
+        return try {
+            mavenArtifactRepository.save(
+                MavenArtifactEntity(
+                    groupId = coords.groupId,
+                    artifactId = coords.artifactId,
+                    version = coords.version,
+                )
+            )
+        } catch (_: DataIntegrityViolationException) {
+            // A concurrent transaction inserted the same GAV first — re-read it.
```
nikitavlaevjb: Do I understand correctly that there is no test for it?
Is it possible to create one?

nikitavlaevjb: Also not sure that after exception it wouldn't just be a rollback

app/src/main/resources/db/migration/2026-Q2/2026-04-23_create_maven_artifact.yml
dkrasnoff: Our ORM layer (hibernate) is using sequences to create IDs, it is made for simplicity on the application layer to get a batch of IDs in advance.
On the other hand, we need to prefill this table and then move the sequence to the actual value to avoid failures.

> Will such backfill advance the id sequence?

No, the initial value for the sequence is set in `create maven_artifact sequence` liquibase changeSet.


> Why not take actual package_id here?

There are maven artifacts, that are not packages in our system + it seems to be better to use a dedicated sequence in terms of the spring layer

app/src/main/resources/db/migration/2026-Q2/2026-04-23_create_maven_artifact.yml
dkrasnoff: Not clear what do you mean by source of truth, but one or more `maven_artifact` should always be present for `package`, at the same time `package` can not be present fro `maven_artifact`.
But your question about cascade delete made me think, and I'm not totally sure we need this constraint exactly.

core/package/src/main/kotlin/io/klibs/core/pckg/service/MavenArtifactService.kt
dkrasnoff: Good questions, I'll check.
I thought that DataIntegrityViolationException is not a RuntimeException, and as I recall only RuntimeException  raises the rollback

app/src/main/resources/db/migration/2026-Q2/2026-04-23_create_maven_artifact.yml
dkrasnoff: Seems to be nothing harmful in having this constraint for now, so I'll leave it just as additional check

core/package/src/main/kotlin/io/klibs/core/pckg/service/MavenArtifactService.kt
dkrasnoff: You were right about the transaction and rollback, I've replaced it with ON CONFLICT DO NOTHING

PR #299, opened 2026-05-05
KTL-4370: Fallback to direct Maven Central upstream if 404

nikitavlaevjb:
Previously, we introduced using cache redirector to decrease maven central requests quota usage.
However, turns out that on cache miss somewhere in the redirector chain (cache-redirector → cloudfront → origin-pull) we get 404, often for new artifacts.
Fallbacking to origin fixes the problem.

### DIFF
--- a/app/src/main/kotlin/io/klibs/app/job/ProcessPackageIndexRequestJob.kt
+++ b/app/src/main/kotlin/io/klibs/app/job/ProcessPackageIndexRequestJob.kt
@@ -9,6 +9,11 @@ import org.springframework.stereotype.Component
 import java.util.concurrent.TimeUnit
 
 @Component
+@ConditionalOnProperty(
+    value = ["klibs.scheduling.process-queue-job.enabled"],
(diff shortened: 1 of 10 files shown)


zwiora (approved):
Looks good to me.

Just one note, I've noticed in the original version `getContentUrlPrefix()` returned `https://search.maven.org/remotecontent?filepath=` and now the `getContentFallbackUrlPrefix` returns `https://repo1.maven.org/maven2/`. I wanted to make sure that's intended.

nikitavlaevjb:
> Looks good to me.
> 
> Just one note, I've noticed in the original version `getContentUrlPrefix()` returned `https://search.maven.org/remotecontent?filepath=` and now the `getContentFallbackUrlPrefix` returns `https://repo1.maven.org/maven2/`. I wanted to make sure that's intended.

Nice catch! I forgot that `https://search.maven.org/remotecontent?filepath=` has slightly more permissive rate limits.

PR #311, opened 2026-06-02
KTL-4519 GitHub repositories on klibs.io doesn't update properly

nikitavlaevjb:
Moved scm_repo backoff mechanism to table.

No need to touch other job's backoff mechanisms, selection strategy there allows us to avoid stagnating on 3 poisoned items.
The proper solution would be present once we introduce queues.

### DIFF
--- a/app/src/main/kotlin/io/klibs/app/util/BackoffProvider.kt
+++ b/app/src/main/kotlin/io/klibs/app/util/BackoffProvider.kt
@@ -6,6 +6,7 @@ import org.slf4j.LoggerFactory
 import java.time.Duration
 import java.time.Instant
 import java.util.concurrent.ConcurrentHashMap
+import kotlin.math.pow
 
 class BackoffProvider(
     private val jobName: String,
@@ -34,7 +35,7 @@ class BackoffProvider(
     fun onFailure(id: Int, now: Instant = Instant.now()) {
         val previous = backoffStates[id]
         val attempts = (previous?.attempts ?: 0) + 1
-        val delay = computeBackoffDelay(attempts)
+        val delay = computeBackoffDelay(attempts = attempts)
         val nextAllowedAt = now.plus(delay)
         backoffStates[id] = BackoffState(attempts, nextAllowedAt)
         logger.warn("Job $jobName: Backoff for id=$id attempts=$attempts nextAllowedAt=$nextAllowedAt delay=${delay.toSeconds()}s")
@@ -44,16 +45,14 @@ class BackoffProvider(
         backoffStates.remove(id)
     }
 
-    private fun computeBackoffDelay(attempts: Int): Duration {
-        // Base 60s, exponential factor 2, capped at 1 hour
-        val base = 60L
-        val exp = 1L shl (attempts - 1).coerceAtMost(10) // cap growth
-        val seconds = (base * exp).coerceAtMost(3600L)
-        return Duration.ofSeconds(seconds)
-    }
-
     companion object {
         private val logger = LoggerFactory.getLogger(BackoffProvider::class.java)
         const val METRIC_NAME = "klibs.backoff.states.size"
+
+        fun computeBackoffDelay(base: Long = 60L, exp: Long = 2L, attempts: Int): Duration {
+            val exp = exp.toDouble().pow(attempts - 1).toLong()
+            val seconds = (base * exp).coerceAtMost(60 * 60 * 24 * 30L) // 1 month
+            return Duration.ofSeconds(seconds)
+        }
     }
 }
--- a/app/src/main/resources/application.yml
+++ b/app/src/main/resources/application.yml
@@ -78,6 +78,8 @@ klibs:
   scheduling:
     process-indexing-queue:
       enabled: ${PROCESS_INDEXING_QUEUE_ENABLED:true}
+    fill-version-type:
+      enabled: ${FILL_VERSION_TYPE_ENABLED:true}
   indexing-configuration:
     executor:
       thread-count: 2
--- a/app/src/test/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryDbTest.kt
+++ b/app/src/test/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryDbTest.kt
@@ -1,0 +1,53 @@
+package io.klibs.core.scm.repository
+
+import BaseUnitWithDbLayerTest
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.test.context.jdbc.Sql
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+/**
+ * Integration tests for [ScmRepositoryRepository.findMultipleForUpdate] and the per-repo backoff
+ * helpers against a Testcontainer Postgres. Seeds repos A–F (see seed-repos.sql); eligible ones,
+ * highest-star first, are D(60), A(50), B(30), F(20). C is in an active backoff window and E was
+ * updated within 24h.
+ */
+class ScmRepositoryRepositoryDbTest : BaseUnitWithDbLayerTest() {
+
+    @Autowired
+    private lateinit var repo: ScmRepositoryRepository
+
+    @Autowired
+    private lateinit var scheduling: ScmRepositorySchedulingRepository
+
+    @Test
+    @Sql("classpath:sql/ScmRepositoryRepositoryDbTest/seed-repos.sql")
+    fun `findMultipleForUpdate skips backed-off and fresh repos, ordered by stars desc`() {
+        val ids = repo.findMultipleForUpdate(10).map { it.idNotNull }
+
+        assertEquals(listOf(800013, 800010, 800011, 800015), ids) // D, A, B, F by stars desc
+    }
+
+    @Test
+    @Sql("classpath:sql/ScmRepositoryRepositoryDbTest/seed-repos.sql")
+    fun `scheduleNextRetry backs the repo off out of the next selection`() {
+        assertTrue(repo.findMultipleForUpdate(10).any { it.idNotNull == 800010 })
+
+        scheduling.scheduleNextRetry(800010, attempts = 1, backoffDelaySeconds = 3600) // next_retry_at an hour ahead
+
+        assertFalse(repo.findMultipleForUpdate(10).any { it.idNotNull == 800010 })
+    }
+
+    @Test
+    @Sql("classpath:sql/ScmRepositoryRepositoryDbTest/seed-repos.sql")
+    fun `clearSchedule clears backoff so a backed-off repo is selectable again`() {
+        // C (800012) is in an active backoff window (next_retry_at in the future).
+        assertFalse(repo.findMultipleForUpdate(10).any { it.idNotNull == 800012 })
+
+        scheduling.clearSchedule(800012) // removes the scheduling row
+
+        assertTrue(repo.findMultipleForUpdate(10).any { it.idNotNull == 800012 })
+    }
+}
--- a/core/scm-repository/src/main/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepository.kt
+++ b/core/scm-repository/src/main/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepository.kt
@@ -18,6 +18,10 @@ interface ScmRepositoryRepository {
 
     fun findIdByName(ownerLogin: String, name: String): Int?
 
+    /**
+     * Next [limit] repos due for a GitHub refresh, highest-star first.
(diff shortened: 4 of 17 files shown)


core/scm-repository/src/main/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryJdbc.kt
```diff
@@ -298,17 +298,42 @@ class ScmRepositoryRepositoryJdbc(
...
+    }
+
+    override fun recordUpdateFailure(id: Int) {
+        // Exponential backoff mirroring BackoffProvider.computeBackoffDelay: base 60s, doubling per
+        // attempt, capped at 1h. The pre-increment retry_attempts is the exponent (so the first
+        // failure waits 60s), and the exponent is capped at 10 to bound the doubling.
+        val sql = """
+            UPDATE scm_repo
+            SET retry_attempts = retry_attempts + 1,
+                next_retry_at = current_timestamp +
```
dkrasnoff: The current formula seems to be calling update very often and doesn't check anything in the future.
WDYT about not having bounds and increasing the power to 3?
Like, `(60 * power(3, retry_attempts)) * interval '1 second' `.

dkrasnoff: If you remove 'parked' logic, please don't forget to remove it from all the comments you left

app/src/main/resources/application.yml
```diff
@@ -74,6 +74,7 @@ klibs:
         request-cache-path: ${KLIBS_CACHE_DIR_REQUESTS:${user.dir}/cache/request-cache}
         request-cache-size-mb: ${KLIBS_CACHE_DIR_REQUEST_SIZE_MB:10}
       update-repos-per-iteration: ${KLIBS_UPDATE_GITHUB_REPOS_PER_ITERATION:3}
+      update-retry-cap: ${KLIBS_UPDATE_GITHUB_RETRY_CAP:5}
```
dkrasnoff: I also think it is not necessary, with a bigger base of a power, a new update will happen rarer

core/scm-repository/src/main/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryJdbc.kt
```diff
@@ -298,17 +298,42 @@ class ScmRepositoryRepositoryJdbc(
...
+                    (least(3600, 60 * power(2, least(retry_attempts, 10))) * interval '1 second')
```
dkrasnoff: It is also kind of business logic, so better to do in on service layer

dkrasnoff (changes requested):

core/scm-repository/src/main/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryJdbc.kt
nikitavlaevjb: Wdym? You want only to move calculation to the service side OR the whole backoff mechanism should not be present in db?

dkrasnoff: Only time calculation

app/src/main/resources/db/migration/2026-Q2/2026-06-02_add_scm_repo_retry_columns.yml
```diff
@@ -0,0 +1,26 @@
+databaseChangeLog:
+  - changeSet:
+      id: add retry columns to scm_repo
+      author: nikita.vlaev@jetbrains.com
+      comment: "KTL-4519: per-repo backoff so failing repos don't occupy every update slot and starve the catalog."
+      preConditions:
+        - onFail: MARK_RAN
```
dkrasnoff: You check only one column and create 2, is it expected to be so?

app/src/main/resources/application.yml
dkrasnoff: remove it

dkrasnoff (changes requested):

app/src/main/kotlin/io/klibs/app/util/BackoffProvider.kt
```diff
@@ -44,16 +45,14 @@ class BackoffProvider(
...
-        return Duration.ofSeconds(seconds)
-    }
-
     companion object {
         private val logger = LoggerFactory.getLogger(BackoffProvider::class.java)
         const val METRIC_NAME = "klibs.backoff.states.size"
+
+        fun computeBackoffDelay(base: Long = 60L, exp: Long = 2L, attempts: Int): Duration {
+            val exp = exp.toDouble().pow(attempts - 1).toLong()
+            val seconds = (base * exp).coerceAtMost(60 * 60 * 24 * 30L) // 1 month
```
dkrasnoff, line 54: nit: make max in 1 year?

app/src/test/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryDbTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        assertTrue(repo.findMultipleForUpdate(10).any { it.idNotNull == 800010 })
+
+        repo.recordUpdateFailure(800010, backoffDelay = 3600) // pushes next_retry_at an hour ahead
+
+        assertFalse(repo.findMultipleForUpdate(10).any { it.idNotNull == 800010 })
+    }
+
+    @Test
+    @Sql("classpath:sql/ScmRepositoryRepositoryDbTest/seed-repos.sql")
+    fun `recordUpdateSuccess clears backoff so a backed-off repo is selectable again`() {
```
nikitavlaevjb: But backoff can be up to 1 month, so it should be findable again after 24h, not in backoff.

app/src/main/kotlin/io/klibs/app/util/BackoffProvider.kt
nikitavlaevjb: It does not matter at this point, or?

app/src/main/resources/db/migration/2026-Q2/2026-06-02_add_scm_repo_retry_columns.yml
nikitavlaevjb: We use `next_retry_at` to check backoff.
We use 'retry_attempts` to calculate backoff.

dkrasnoff: In the `preConditions` block of this Liquibase script, you check only the existence of `next_retry_at`, but create two columns, `retry_attempts` and `next_retry_at`, in one script. Is it expected?

nikitavlaevjb: Oh, that's wrong, thanks!

app/src/test/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryDbTest.kt
dkrasnoff: If we put the record in backoff, we cannot update it for some reason, so we most probably will not update it in 24h either. I'd say backoff is the source of truth in our case.

Additional question:
Why then do we have two different fields for the same thing?

app/src/main/kotlin/io/klibs/app/util/BackoffProvider.kt
dkrasnoff: Hm, rethought my message, 1month is good

app/src/test/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryDbTest.kt
nikitavlaevjb: This is the test case for when we tried after the backoff and succeeded.
Then the backoff should be cleared.

> Why then do we have two different fields for the same thing?

It would be confusing if we would put `updatedAtTs` in the future to emulate backoff, if that's what you're proposing

core/scm-repository/src/main/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepository.kt
```diff
@@ -18,6 +18,16 @@ interface ScmRepositoryRepository {
...
     fun findIdByName(ownerLogin: String, name: String): Int?
 
+    /**
+     * Next [limit] repos due for a GitHub refresh, highest-star first.
+     * Skips repos still within their backoff window (`next_retry_at`)
+     */
     fun findMultipleForUpdate(limit: Int = 3): List<ScmRepositoryEntity>
+
+    /** Clears backoff state after a successful update so future failures start fresh. */
+    fun recordUpdateSuccess(id: Int)
```
nikitavlaevjb: No, I like that it is structured outside of it, symmetric to recordUpdateFailure

app/src/main/resources/application.yml
nikitavlaevjb: fixed

app/src/main/resources/db/migration/2026-Q2/2026-06-02_add_scm_repo_retry_columns.yml
nikitavlaevjb: fixed

app/src/test/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryDbTest.kt
dkrasnoff: > It would be confusing if we would put updatedAtTs in the future to emulate backoff, if that's what you're proposing

It is what I'm proposing. What will be confusing in it? Could you elaborate?

nikitavlaevjb: to make it logical, we would have to rename `updatedAtTs` to `nextUpdateTs` and rewire some conditions around it.
If we do so, we would lose the actual `updatedAtTs` history data, which is sometimes needed for investigations

core/scm-repository/src/main/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepository.kt
dkrasnoff: Could you elaborate on the pros of symmetry in this case, please?

app/src/test/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryDbTest.kt
dkrasnoff: I see your concern now. Thanks for the clarification. Yeah, let's not merge it into one field.

At the same time, the test still shows that the current solution could lead to mistakes in the future when we update the backoff but don't update `updatedAtTs` on this DTO, and vice versa.
So, my suggestion is to update it in one place.

core/scm-repository/src/main/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepository.kt
nikitavlaevjb: 1. `updateRepo` is already overloaded method
2. `syncRepositoryWithGitHub` method handles errors, and backoffs for that reason
So here is a win by separation of concerns.

Another thing: if we move backoff handling responsibility to `ScmRepositoryRepository.update`, then every `update` call would be responsible for thinking about backoffs -- I don't want that.

app/src/test/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepositoryDbTest.kt
nikitavlaevjb: I think that `updatedAtTs` and `nextRetryAt` are supposed to be update separately because they represent different things.
`updatedAtTs` is about data, and is set even in case of "logical" failure ` if (ghRepo == null) {`.
`nextRetryAt` is about scheduling, and it should be a responsibility of a job that relies on it.

dkrasnoff: discussed in DM, Nikita will normalize the `nextRetryAt` and make a separate table for it

core/scm-repository/src/main/kotlin/io/klibs/core/scm/repository/ScmRepositoryRepository.kt
dkrasnoff: discussed in DM, Nikita will normalize the `nextRetryAt` and make a separate table for it

dkrasnoff (approved):

PR #331, opened 2026-07-06
Ktl 4552 add ai page

krutilov:


frontend/src/app/ai/page.tsx
```diff
@@ -0,0 +1,152 @@
+"use client"
```
nikpachoo, line 1: Maybe we don't need it?

nikpachoo (approved):

PR #333, opened 2026-07-08
Removed featured category

dkrasnoff:


app/src/main/resources/db/migration/2026-Q3/2026-07-08_remove-featured-category.yml
```diff
@@ -0,0 +1,30 @@
+databaseChangeLog:
+  - changeSet:
+      id: remove featured category
+      author: dmitrii.krasnov@jetbrains.com
+      changes:
+        - delete:
+            tableName: category
+            where: name = 'Featured'
+  - changeSet:
```
nikitavlaevjb, line 9: We will still have gaps in category ids, is that ok?

core/project/src/main/kotlin/io/klibs/core/project/enums/MarkerType.kt
```diff
@@ -4,7 +4,8 @@ package io.klibs.core.project.enums
  * This enum encounters all known project awards (see KTL-2318)
  */
 enum class MarkerType {
-    FEATURED,
+    FEATURED, // TODO(Dmitrii Krasnov): should be removed in two steps, so remove it, when KTL-4602 will be on prod
```
nikitavlaevjb, line 7: What about "Featured" label on frontend? (blue project card with a rocket)

app/src/test/kotlin/io/klibs/core/search/CategoriesProjectsDbTest.kt
```diff
@@ -54,9 +54,9 @@ class CategoriesProjectsDbTest : BaseUnitWithDbLayerTest() {
     fun `limit is respected per category`() {
         val results = searchService.searchByCategories(limit = 2)
 
-        val featured = results.first { it.categoryName == "Featured" }
-        assertEquals(2, featured.projects.size, "FEATURED capped at limit=2")
-        assertEquals(listOf(100, 90), featured.projects.map { it.vcsStars }, "Top-2 by stars")
+        val featured = results.first { it.categoryName == "Grant winners" }
```
nikitavlaevjb, line 57: rename variable also.
Run agent to look for similar kind of mistakes too?

nikitavlaevjb (changes requested):

core/project/src/main/kotlin/io/klibs/core/project/enums/MarkerType.kt
dkrasnoff: `FEATURED` is saved only for backward compatibility during deployment and caching.
After all caches are renewed, there will be no FEATURED returned to the frontend, so nothing will be blue.

app/src/main/resources/db/migration/2026-Q3/2026-07-08_remove-featured-category.yml
dkrasnoff: I thought about it, and it seems to be ok, because they are used only for ordering, and ordering is fine now.

app/src/test/kotlin/io/klibs/core/search/CategoriesProjectsDbTest.kt
dkrasnoff: Thanks!
I didn't run the agent for this feature, so I could skip such renamings. Will fix and run agent to check for more places :)

core/project/src/main/kotlin/io/klibs/core/project/enums/MarkerType.kt
nikitavlaevjb: So you want to keep the stale frontend logic too for now?

dkrasnoff: Yes, I also think it could be reused in `trending` task

PR #337, opened 2026-07-09
KTL-4691: Removed OSS Health: surface-level

nikitavlaevjb:
Separated this PR from the full removal to integrate it faster.
Full removal is incoming soon.


dkrasnoff:
Discussed with Nikita that the backend calculation does not need to be removed right now.
So, from my end is ok.

dkrasnoff (approved):

krutilov (approved):

PR #340, opened 2026-07-09
KTL-4673: decouple description regen from reindex

nikitavlaevjb:


app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -213,20 +217,23 @@ class PackageIndexingService(
         mavenArtifact: MavenArtifact,
         pom: MavenPom,
         toolingMetadata: KotlinToolingMetadataDelegate,
-        projectEntity: ProjectEntity?
+        projectEntity: ProjectEntity?,
+        reindex: Boolean
     ): PackageDTO {
-        val (description, descriptionWasGenerated) = resolvePackageDescription(pom)
+        // should be backfilled before if null
```
dkrasnoff: nit: this comment has no sense, because the check below does the same.

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -235,38 +242,65 @@ class PackageIndexingService(
...
 
-        // If a previous version exists and had a generated description, generate a new description
-        var description = pom.description
-        var descriptionWasGenerated = false
+        // Regenerate only for a genuinely new latest version whose predecessor had a generated
+        // description. A reindex (deps/targets backfill) or an older/equal version must never
+        // trigger web-search regeneration (KTL-4673).
+        if (reindex ||
+            previousVersion == null ||
+            !releasedAt.isAfter(previousVersion.releaseTs) ||
```
dkrasnoff: Why do we need a check that previous version is before this version?
If it still possible, then I'd rename it to `latestSavedVersion` and add code like this:
```
        val latestSavedVersion = packageRepository.findFirstByGroupIdAndArtifactIdOrderByReleaseTsDesc(
            pom.groupId, pom.artifactId
        )
        val previousVersion = latestSavedVersion?.takeIf { releasedAt.isAfter(it.releaseTs) }

        // Regenerate only for a genuinely new latest version whose predecessor had a generated
        // description. A reindex (deps/targets backfill) or an older/equal version must never
        // trigger web-search regeneration (KTL-4673).
        if (reindex ||
            previousVersion == null ||
            !previousVersion.generatedDescription
        ) {
            return ResolvedDescription(pom.description, wasGenerated = false, generatedAt = null)
        }
```

dkrasnoff: How is it possible, that `!releasedAt.isAfter(previousVersion.releaseTs)` could be false?

app/src/test/kotlin/io/klibs/app/indexing/PackageIndexingServiceTest.kt
```diff
@@ -287,6 +296,112 @@ class PackageIndexingServiceTest : BaseUnitWithDbLayerTest() {
...
+        val generatedAtBefore = before.descriptionGeneratedAt
+        assertNotNull(generatedAtBefore, "Precondition: seeded package has a generated timestamp")
+
+        // If the AI generator were ever invoked its result would be this sentinel; it must never reach the DB.
+        whenever(packageDescriptionGenerator.generatePackageDescription(any(), any(), any(), any(), any()))
+            .thenReturn("AI SENTINEL - must not be persisted on reindex")
+
+        stubMavenFetch(groupId, artifactId, version, pomDescription = "Fresh POM description")
+
+        assertTrue(uut.processPackageQueue(), "Should process the reindex request")
```
dkrasnoff: nit: we can use `uut.processRequest` instead of `uut.processPackageQueue()` and in the next test as well.

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
nikitavlaevjb: If you mean that it seems duplicate to the 'reindex' check, it almost is.
But there is a rare case when we actually miss an older version of a package for any reason, like incomplete index or our bug.
In that case, we dont want to generate description either.

Otherwise your suggestion looks fine, but this line lacks some common sense for me
```
val previousVersion = latestSavedVersion?.takeIf { releasedAt.isAfter(it.releaseTs) }
```
Like why would a `previousVersion` be null if it is not older than the incoming one?

nikitavlaevjb: See above

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
nikitavlaevjb: I added it myself because for me it is not obvious why we `require` this here looking only at this function.

dkrasnoff: I mean for me it is still not obvious why and how I should backfill, it just say that I should, but I will understand it also from the failed require.

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
dkrasnoff: I'd like to decouple this topic, because for me it is not obvious how it is possible, so I'll just continue here, if you don't mind.

> But there is a rare case when we actually miss an older version of a package for any reason, like incomplete index or our bug.

Do you mean that `releasedAt` for current package can be earlier that `previousVersion.releaseTs` even if `reindex == false`?

nikitavlaevjb: Yes, exactly. For instance, if we deleted the "bugged" versions from index_request and they got discovered again

dkrasnoff: I just wanted to make naming a little bit more convinient. 
In your case, I was confused, how is it possible to have `previousVersion` with later release date, but actually it was not the `previousVersion`, but latest saved version to our database. 
So, my point here only about naming. In my case `previousVersion` is null when the `latestSavedVersion` `is not older than the incoming one`, because we didn't find `previousVersion` - so it is either bug in our algorithm or it doesn't exist.

nikitavlaevjb: Ok, makes sense to rename `previousVersion` to `latestSavedVersion` indeed!

dkrasnoff: Great, could you leave a comment with example then

app/src/test/kotlin/io/klibs/app/indexing/PackageIndexingServiceTest.kt
nikitavlaevjb: fixed

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
nikitavlaevjb: fixed

nikitavlaevjb: fixed

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
nikitavlaevjb: removed

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
```diff
@@ -259,29 +258,30 @@ class PackageIndexingService(
...
 
         // Regenerate only for a genuinely new latest version whose predecessor had a generated
         // description. A reindex (deps/targets backfill) or an older/equal version must never
         // trigger web-search regeneration (KTL-4673).
         if (reindex ||
-            previousVersion == null ||
-            !releasedAt.isAfter(previousVersion.releaseTs) ||
-            !previousVersion.generatedDescription
+            latestSavedVersion == null ||
+            // A fresh version can be older than last saved: e.g. because of bug the fix in our backend
```
dkrasnoff: nit: replace to  "fresh version can be older, for instance, if we deleted the "bugged" versions from index_request and they got discovered again"

dkrasnoff (approved):

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
nikitavlaevjb: fixed

PR #341, opened 2026-07-10
KTL-4062: Added index package from GitHub requests

dkrasnoff:


dkrasnoff:
I bypass rules, because I was reviewer of this PR and it was mainly developed by Zofia.
So, no need in additional review.

PR #344, opened 2026-07-16
KTL-4226 Improve downloading Maven Index

zwiora:
Setting `updateRequest.localIndexCacheDir` should make the `indexUpdater.fetchAndUpdateIndex(updateRequest)` first save the zip in provided directory and only later extract it

### DIFF
--- a/integrations/maven/src/main/kotlin/io/klibs/integration/maven/service/MavenIndexingContextManager.kt
+++ b/integrations/maven/src/main/kotlin/io/klibs/integration/maven/service/MavenIndexingContextManager.kt
@@ -34,6 +34,8 @@ class MavenIndexingContextManager(
 
     fun getIndexTmpDir(): File = File("${properties.central.indexDir}/tmp")
 
+    fun getLocalIndexCacheDir(): File = File("${properties.central.indexDir}/local-index-cache")
+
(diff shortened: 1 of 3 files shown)


integrations/maven/src/main/kotlin/io/klibs/integration/maven/service/MavenIndexingContextManager.kt
```diff
@@ -34,6 +34,8 @@ class MavenIndexingContextManager(
 
     fun getIndexTmpDir(): File = File("${properties.central.indexDir}/tmp")
 
+    fun getLocalIndexCacheDir(): File = File("${properties.central.indexDir}/local-index-cache")
```
dkrasnoff, line 37: Should we maintain clearing it ourselves? Or will it be automatically cleared?

dkrasnoff: And should the subdir `local-index-cache` be created in advance, or is it also handled under the hood?

zwiora: The subdir will be automatically created and the zip will be cleared at the beginning of next run

dkrasnoff (approved):

PR #362, opened 2026-07-24
KTL-4744: Fix sticky search bar ignoring typing and losing focus

ivan-orekhov:
## Problem

[KTL-4744](https://youtrack.jetbrains.com/issue/KTL-4744)

The search page renders two `SearchField` instances — the main one in the hero block and a compact one in the sticky bar shown after scrolling past the main filter. In non-category mode the compact instance was wired with `onChange={undefined}`, so typing there never triggered the search and only Enter worked; to users this looked like an intermittent "search ignores typing" bug, depending on which of the two visually identical fields they typed in.

When a search did fire while scrolled down, the results swap collapsed the page height, the browser clamped the scroll to top, the sticky bar unmounted mid-typing, and focus (plus un-flushed debounced text) was lost.

## Changes

- **`SearchField` is now fully controlled**: the query text lives in the parent, so both bars always render the same text and none of it is lost when one of them unmounts. Debouncing moves to the callers via a new shared `useDebouncedCallback` hook (invokes the latest callback, exposes `cancel()`).
- **Compact bar triggers search-as-you-type** — wired to the same handler as the main field.
- **Focus handoff**: when results replace the content and the page ends up at the top, the sticky bar hands focus (caret at the end) and text over to the main field, so typing continues seamlessly.
- **Scroll behavior**: typing-driven URL updates use `scroll: false`; sort/tag/platform/mode changes keep the default scroll-to-top. Enter commits the search immediately (cancels the pending debounce). Escape/Clear on an empty field stays a no-op.
- **Navbar**: search input got its own controlled state; suggestion fetches are debounced, skipped for an empty query, and cancelled on Enter navigation.
- **Test**: Playwright regression test — typing in the compact bar triggers the search without Enter, and the focus handoff preserves the text.

## Testing

- `npx tsc --noEmit` and `next lint` clean.
- `tests/search.spec.ts`: 8/8 passing (incl. the new regression test) against a local dev server with the production API.
- Full Playwright suite: 72 passed; the 8 failures in `category-page`/`package_page`/`project_page_labels` are pre-existing on a clean checkout (they expect the internal test API dataset, which is VPN-only).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


frontend/src/app/ui/navbar/index.tsx
```diff
@@ -43,21 +44,33 @@ export default function Navbar() {
...
     // Dropdown menu constants
     const [isOpen, setIsOpen] = useState(false);
     const toggleIsOpen = () => setIsOpen(s => !s);
 
     // Suggestions. For now we're using search query with limit
+    const [searchQuery, setSearchQuery] = useState("");
     const [searchSuggestions, setSearchSuggestions] = useState<ProjectSearchResults[] | null>(null)
 
-    const handleNavbarSearchInput = useCallback(async (value: string) => {
+    const fetchSuggestions = useDebouncedCallback(async (value: string) => {
```
krutilov: This is some leftover logic from suggestion function that we decided to disable some ago and eventually deprecated. So would be nice to remove all it's logic

krutilov (changes requested):

PR #369, opened 2026-07-29
KTL-4711 OpenSearch query repositories

nikitavlaevjb:
- OpenSearchTempPopulator is a temporary solution; will get rid of it once we add a proper sync pipeline
- The updated headline score is 95/107, headline 0.7334. Will freeze it after we finish with the migration.


core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexBootstrap.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
+import java.io.StringReader
+
+/**
+ * Creates the project + package indices with their mappings on startup if absent.
+ */
+@Component
+@ConditionalOnProperty("klibs.search.opensearch.enabled", havingValue = "true")
+class IndexBootstrap(
```
dkrasnoff: Not clear from name which index we are bootstapping. It is better to explicitly highlight it in the name

core/search/src/main/resources/opensearch/project-mappings.json
```diff
@@ -0,0 +1,32 @@
+{
+  "properties": {
+    "project_id": { "type": "long" },
```
dkrasnoff, line 3: should we add `health_score` as we use it in OpenSearchTempPopulator and ProjectSearchRepositoryOpenSearch ?

core/search/src/main/resources/opensearch/project-mappings.json
```diff
@@ -0,0 +1,32 @@
...
+    "platforms": { "type": "keyword" },
+    "tags": { "type": "text", "fields": { "keyword": { "type": "keyword" } } },
+    "markers": { "type": "keyword" },
+    "targets": { "type": "keyword" },
+    "stars": { "type": "integer" },
+    "dependent_count": { "type": "integer" },
+    "has_readme": { "type": "boolean" },
+    "license_name": { "type": "keyword" },
+    "latest_version": { "type": "keyword" },
+    "latest_version_ts": { "type": "keyword" },
```
dkrasnoff: why not date ?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchTempPopulator.kt
```diff
@@ -0,0 +1,163 @@
...
+    fun populatePackages() = rebuild(properties.packageIndex, PACKAGE_DOC_SQL) { node ->
+        "${node.get("group_id").asText()}:${node.get("artifact_id").asText()}"
+    }
+
+    private fun rebuild(index: String, sql: String, idField: String) =
+        rebuild(index, sql) { it.get(idField).asText() }
+
+    private fun rebuild(index: String, sql: String, idOf: (ObjectNode) -> String) {
+        client.deleteByQuery { d -> d.index(index).query { q -> q.matchAll { it } } }
+        val docs = jdbcClient.sql(sql).query(String::class.java).list()
```
dkrasnoff, line 40: For a prod-size corpus this is a large heap, can we stream row-by-row into the bulk batches ?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchTempPopulator.kt
```diff
@@ -0,0 +1,163 @@
...
+
+    fun populatePackages() = rebuild(properties.packageIndex, PACKAGE_DOC_SQL) { node ->
+        "${node.get("group_id").asText()}:${node.get("artifact_id").asText()}"
+    }
+
+    private fun rebuild(index: String, sql: String, idField: String) =
+        rebuild(index, sql) { it.get(idField).asText() }
+
+    private fun rebuild(index: String, sql: String, idOf: (ObjectNode) -> String) {
+        client.deleteByQuery { d -> d.index(index).query { q -> q.matchAll { it } } }
```
dkrasnoff, line 39: A mid-refill failure leaves the index empty/partial, so a failed populate silently yields empty search results.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchConfiguration.kt
```diff
@@ -0,0 +1,44 @@
+package io.klibs.core.search.opensearch
```
dkrasnoff: move to io.klibs.core.search.configuration please

core/search/src/main/kotlin/io/klibs/core/search/configuration/properties/OpenSearchProperties.kt
```diff
@@ -0,0 +1,17 @@
+package io.klibs.core.search.opensearch
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+/**
+ * OpenSearch connection + index config (KTL-4711). v1 = lexical only; semantic/hybrid knobs
+ * are Phase 2. `enabled` gates the whole OS stack off for normal app boot / non-OS tests.
+ */
+@ConfigurationProperties("klibs.search.opensearch")
+data class OpenSearchProperties(
```
dkrasnoff, line 10: move to io.klibs.core.search.configuration.properties please

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchQueryBuilder.kt
```diff
@@ -0,0 +1,140 @@
...
+            .boost(boost.toFloat())
+            .build()
+            .toQuery()
+
+    fun term(field: String, value: String): Query =
+        Query.of { q -> q.term { t -> t.field(field).value(FieldValue.of(value)) } }
+
+    // OR between values for a field
+    fun termsAny(field: String, values: List<String>): Query {
+        // Wwe want to achieve something like:
```
dkrasnoff: typo

gradle/libs.versions.toml
```diff
@@ -26,6 +26,9 @@ caffeine = "3.2.3"
 maven-model = "3.9.9"
 maven-indexer = "7.1.6"
 
+# Search engine (KTL-2933 / KTL-4711)
```
dkrasnoff: no need in the links to the tasks

core/search/src/test/kotlin/io/klibs/core/search/opensearch/IndexDefinitionsTest.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.jupiter.api.Test
+import org.opensearch.client.json.JsonpDeserializer
+import org.opensearch.client.json.jackson.JacksonJsonpMapper
+import org.opensearch.client.opensearch._types.mapping.TypeMapping
+import org.opensearch.client.opensearch.indices.IndexSettings
+import java.io.StringReader
+
+class IndexDefinitionsTest {
+
+    private val mapper = JacksonJsonpMapper()
```
dkrasnoff: Can we reuse mapper from the io.klibs.core.search.opensearch.OpenSearchConfiguration#openSearchClient ? 
I mean create the original one as a separate bean and use this bean here, so that when we change the original one, it will be the same here

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchTempPopulator.kt
nikitavlaevjb: > OpenSearchTempPopulator is a temporary solution; will get rid of it once we add a proper sync pipeline

Just ignore this class

core/search/src/main/resources/opensearch/project-mappings.json
nikitavlaevjb: Thats an unused artifact, forgot to remove it, thanks

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexBootstrap.kt
nikitavlaevjb: This will be also gone once we will ad a sync pipeline.
But anyway, I'll fix it

core/search/src/main/resources/opensearch/project-mappings.json
nikitavlaevjb: API compatibility on the repository level.
We can move this compatibility to the service level, then we can add a date.
But I don't really see the "why"

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchTempPopulator.kt
nikitavlaevjb: Ignore it plz

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchTempPopulator.kt
dkrasnoff: sorry I saw your initial comment, but forgot during the reveiw

core/search/src/test/kotlin/io/klibs/core/search/opensearch/IndexDefinitionsTest.kt
```diff
@@ -0,0 +1,31 @@
...
+    @Test
+    fun `resources load and parse into typed settings and mappings`() {
+        assertNotNull(parse(IndexDefinitions.PROJECT_SETTINGS, IndexSettings._DESERIALIZER))
+        assertNotNull(parse(IndexDefinitions.PACKAGE_SETTINGS, IndexSettings._DESERIALIZER))
+
+        val projectMappings = parse(IndexDefinitions.PROJECT_MAPPINGS, TypeMapping._DESERIALIZER)
+        val packageMappings = parse(IndexDefinitions.PACKAGE_MAPPINGS, TypeMapping._DESERIALIZER)
+
+        assertTrue(projectMappings.properties().containsKey("packages"))
+        assertTrue(projectMappings.properties()["project_id"]!!.isLong)
```
dkrasnoff: Not clear from me should this test check the whole structure or only these two properties?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
```diff
@@ -0,0 +1,52 @@
...
+@Component
+@ConditionalOnProperty("klibs.search.opensearch.enabled", havingValue = "true")
+class IndexBootstrap(
+    private val client: OpenSearchClient,
+    private val properties: OpenSearchProperties,
+) {
+
+    private val mapper = JacksonJsonpMapper()
+
+    @EventListener(ApplicationReadyEvent::class)
```
dkrasnoff, line 30: There could be a small window between application start and this bootstrap, where user requests could come.
Will it be in the final production solution or is it a temporary solution?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
```diff
@@ -0,0 +1,52 @@
...
+        createIfAbsent(properties.projectIndex, IndexDefinitions.PROJECT_SETTINGS, IndexDefinitions.PROJECT_MAPPINGS)
+        createIfAbsent(properties.packageIndex, IndexDefinitions.PACKAGE_SETTINGS, IndexDefinitions.PACKAGE_MAPPINGS)
+    }
+
+    private fun createIfAbsent(index: String, settingsJson: String, mappingsJson: String) {
+        if (client.indices().exists { it.index(index) }.value()) {
+            log.info("OpenSearch index '{}' already exists", index)
+            return
+        }
+        val settings = parse(settingsJson, IndexSettings._DESERIALIZER)
```
dkrasnoff, line 41: should it be a part of infrastructure code instead of a backend application?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    @EventListener(ApplicationReadyEvent::class)
+    fun bootstrap() {
+        createIfAbsent(properties.projectIndex, IndexDefinitions.PROJECT_SETTINGS, IndexDefinitions.PROJECT_MAPPINGS)
+        createIfAbsent(properties.packageIndex, IndexDefinitions.PACKAGE_SETTINGS, IndexDefinitions.PACKAGE_MAPPINGS)
+    }
+
+    private fun createIfAbsent(index: String, settingsJson: String, mappingsJson: String) {
+        if (client.indices().exists { it.index(index) }.value()) {
+            log.info("OpenSearch index '{}' already exists", index)
```
dkrasnoff, line 38: Not clear for me now, how will we update index structure in the future? I don't ask to do it in this task, just want to understand how, is there anything like liquibase but for open search structures or do they have other solutions?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexBootstrap.kt
```diff
@@ -0,0 +1,52 @@
...
+    private val client: OpenSearchClient,
+    private val properties: OpenSearchProperties,
+) {
+
+    private val mapper = JacksonJsonpMapper()
```
dkrasnoff: nit: can we inject it via constructor as other fields?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchConfiguration.kt
```diff
@@ -0,0 +1,44 @@
...
+@EnableConfigurationProperties(OpenSearchProperties::class)
+@ConditionalOnProperty("klibs.search.opensearch.enabled", havingValue = "true")
+class OpenSearchConfiguration {
+
+    @Bean
+    fun openSearchClient(properties: OpenSearchProperties): OpenSearchClient {
+        val uri = URI(properties.uri)
+        val host = HttpHost(uri.scheme, uri.host, uri.port)
+        val transport = ApacheHttpClient5TransportBuilder.builder(host)
+            .setMapper(JacksonJsonpMapper())
```
dkrasnoff: [duplicate just in case] this one could be extracted to a separate bean and used in IndexBootstrap.kt and IndexDefinitionsTest.kt

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
```diff
@@ -0,0 +1,106 @@
...
+                .groupBy(
+                    keySelector = { TargetGroup.fromPlatformAndTarget(it.platform.name, it.target!!) },
+                    valueTransform = { it.target!! },
+                )
+                .mapValues { it.value.toSet() },
+        )
+    }
+
+    private companion object {
+        private val SOURCE_FIELDS = listOf(
```
dkrasnoff: Is it the only way for declaring fields in the codebase? Are there any ways like autogeneration some classes from the given schema?
Now I see that you hardcode "group_id" in 3 different places at least in this file, mybe more in the whole codebase.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchQueryBuilder.kt
```diff
@@ -0,0 +1,140 @@
...
+    // - Stars have 2.5x more weight than Dependent_count
+
+    private const val POPULARITY_SCRIPT =
+        "double d = doc['has_readme'].value ? 1.0 : 0.7; " +
+                "return 1 + (" +
+                "Math.log(doc['stars'].value + 1) * 0.5 + " +
+                "Math.log(doc['dependent_count'].value + 1) * 0.2" +
+                ") * d;"
+
+    // Word-bag match: OR over the query terms, docs matching only some of them still score.
```
dkrasnoff: Didn't get this sentence, could you elaborate?

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
```diff
@@ -0,0 +1,106 @@
...
+            add(fuzzy("group_id", query, 2))
+            add(fuzzy("artifact_id", query, 2))
+            add(phrasePrefix("group_id", query, 2))
+            add(phrasePrefix("artifact_id", query, 2))
+            if (multiWord) add(phrase("artifact_id", query, 4))
+        }
+    }
+
+    override fun sortOptions(sortBy: SearchSort, isQueryPresent: Boolean): List<SortOptions> {
+        val primary = if (sortBy == SearchSort.RELEVANCY && isQueryPresent) {
```
dkrasnoff, line 68: Will this part be modified in the following tasks?
 What about MOST_STARS, MOST_DEPENDENTS?

core/search/src/main/kotlin/io/klibs/core/search/repository/ProjectSearchRepositoryOpenSearch.kt
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    private fun extraFilters(tags: List<String>, markers: List<String>): List<Query> = buildList {
+        tags.forEach { add(OpenSearchQueryBuilder.term("tags.keyword", it)) }
+        if (markers.isNotEmpty()) add(OpenSearchQueryBuilder.termsAny("markers", markers))
+    }
+
+    override fun sortOptions(sortBy: SearchSort, isQueryPresent: Boolean): List<SortOptions> {
+        val primary = when {
```
dkrasnoff: nit: idea suggested this refactoring:
```
        val primary = when (sortBy) {
            SearchSort.RELEVANCY if isQueryPresent -> scoreDesc()
            SearchSort.MOST_DEPENDENTS -> fieldSort("dependent_count", SortOrder.Desc)
            SearchSort.MOST_HEALTHY -> fieldSort("health_score", SortOrder.Desc)
            else -> fieldSort("stars", SortOrder.Desc)
        }
```

dkrasnoff (changes requested):

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
nikitavlaevjb: There are two ways to go: jdbc fallback or health-check way.
I chose the second.
Basically, the pod is not ready until the index is bootstrapped.
In the meantime, the older pod continues serving from the old index.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
nikitavlaevjb: settings.json is for the index functionality, thus a part of application.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
nikitavlaevjb: There are versioned schema files, named `*-mapping.json`.
Since opensearch is not a primary db but only a search index, imo there is no point for migrations

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
nikitavlaevjb: Its actually an optimization and I assume that it is often not the intended way to use the apis.
But I'll look it up, maybe will find something nicer

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchQueryBuilder.kt
nikitavlaevjb: word-bag: order does not matter, nothing matters except the words.
"docs matching only some of them still score": partial matches allowed. 
Example query: "ktor client".
"ktor-server" still matches with low score.

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
nikitavlaevjb: Thats a bug, thx

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
dkrasnoff: Yeah, we'll still add or remove fields from it from time to time. How will it happen?
Could you describe highlevel picture, like:
1) We add new `*-mapping.json`
2) Deploy it to production
3) Some time live with old and new index
4) Switch to new
5) Remove old

nikitavlaevjb: Almost correct, step 3 will exist for a very small amount of time on redeploy.
New pod will not be marked "healthy" until the requested new index has been populated.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexBootstrap.kt
nikitavlaevjb: fixed

core/search/src/main/resources/opensearch/project-mappings.json
nikitavlaevjb: fixed

core/search/src/main/resources/opensearch/project-mappings.json
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchConfiguration.kt
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/configuration/properties/OpenSearchProperties.kt
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchQueryBuilder.kt
nikitavlaevjb: fixed

gradle/libs.versions.toml
nikitavlaevjb: fixed

core/search/src/test/kotlin/io/klibs/core/search/opensearch/IndexDefinitionsTest.kt
nikitavlaevjb: fixed

core/search/src/test/kotlin/io/klibs/core/search/opensearch/IndexDefinitionsTest.kt
nikitavlaevjb: it is just a sanity check on the fields from the mapping that will never be gone from there

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
nikitavlaevjb: changed to "exclude"

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchQueryBuilder.kt
nikitavlaevjb: explained in the comment

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/repository/ProjectSearchRepositoryOpenSearch.kt
nikitavlaevjb: fixed

core/search/src/main/resources/opensearch/project-mappings.json
dkrasnoff: Thanks for changing!
Just to clarify, have you checked that in we didn't missed any error in conversion, for example previous PG stores it in ms, in service layer it become second and we just wrote ms as seconds?

core/search/src/test/kotlin/io/klibs/core/search/opensearch/IndexDefinitionsTest.kt
dkrasnoff: change it to `private val mapper = OpenSearchConfiguration().openSearchJsonpMapper()` please, because openSearchJsonpMapper can change under the hood as well.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
dkrasnoff: Didn't get your answer, why are you sure that "the pod is not ready until the index is bootstrapped"?
If I change this code block to:
```
   fun bootstrap() {
        log.info("Waiting for OpenSearch to be ready")
        sleep(20000)
        log.info("Finished for OpenSearch delay")
        createIfAbsent(properties.projectIndex, IndexDefinitions.PROJECT_SETTINGS, IndexDefinitions.PROJECT_MAPPINGS)
        createIfAbsent(properties.packageIndex, IndexDefinitions.PACKAGE_SETTINGS, IndexDefinitions.PACKAGE_MAPPINGS)
    }
```
I see that `Finished for OpenSearch delay` appears much later than the application initialization finished, and it is ready to get requests.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
dkrasnoff: Got it, thanks!

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
dkrasnoff: Sorry, I meant another thing. You don't need to hardcode each field in different places. It should be avoided

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
dkrasnoff: It seems you didn't change anything in this method. Should I search changes somewhere else?

dkrasnoff (changes requested):

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
nikitavlaevjb: I was talking about the final production solution, as you've mentioned it in the orig message.
This currently is a temporary solution.
Sorry if I did not elaborate on it earlier

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
nikitavlaevjb: I got it, so I minimized the scope of those hardcodes (there is no better solution of this here besides not doing it at all).
We can extract it to constant if you want to.

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
nikitavlaevjb: my bad, fix did not land

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
dkrasnoff: Yeah I expected that they have smth similar to https://docs.hibernate.org/orm/5.3/topical/html_single/metamodelgen/MetamodelGenerator.html.
Could you extract to constants please

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
dkrasnoff: Got it, thanks!
Could we somehow mark it in the code, now it is not obvious

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
nikitavlaevjb: actually, I got what happened.
This is package search, not project search :)
This does not require any stars or dependents sort

core/search/src/main/resources/opensearch/project-mappings.json
nikitavlaevjb: yeah, its never converted to integer, its Instant -> PG Timestamp -> string -> Instant

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
```diff
@@ -0,0 +1,103 @@
...
+import org.opensearch.client.opensearch._types.SortOptions
+import org.opensearch.client.opensearch._types.SortOrder
+import org.opensearch.client.opensearch._types.query_dsl.Query
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.context.annotation.Primary
+import org.springframework.stereotype.Repository
+import java.time.LocalDateTime
+import java.time.ZoneOffset
+
+@Primary
```
dkrasnoff: We don't fall back to PostgreSQL FTS if something goes wrong with the search. Am I right that you will implement it in one of the next tasks?

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
dkrasnoff: Discussed in DM that we will consider "https://github.com/opensearch-project/spring-data-opensearch"

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
dkrasnoff: Leave a comment in project sort and hear about repeatable search sort

core/search/src/test/kotlin/io/klibs/core/search/opensearch/IndexDefinitionsTest.kt
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexBootstrap.kt
nikitavlaevjb: added "TEMPORARY"

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
nikitavlaevjb: done

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
nikitavlaevjb: added a fallback here

dkrasnoff (approved):

PR #384, opened 2026-08-06
KTL-4713 PG→OpenSearch sync pipeline

nikitavlaevjb:
This PingController and SearchIndexReadiness stuff could be avoided if we used actuator instead of pings.


core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexNaming.kt
```diff
@@ -0,0 +1,26 @@
+package io.klibs.core.search.opensearch
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+
+/** The indices this version builds, one [IndexSpec] each. */
```
dkrasnoff: Which version?

app/src/main/kotlin/io/klibs/app/controller/PingController.kt
```diff
@@ -1,10 +1,20 @@
...
 
+import io.klibs.app.search.SearchIndexReadiness
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.RestController
 
 @RestController
-class PingController {
+class PingController(
```
dkrasnoff, line 15: Could you please add documentation to this controller - why do we need it, and why don't we use health endpoint

core/search/src/main/kotlin/io/klibs/core/search/service/SearchService.kt
```diff
@@ -112,19 +110,18 @@ class SearchService(
             }
     }
 
-    /** `project_index` also backs [searchByCategories], so it is refreshed in both engines' modes. */
+    /**
```
dkrasnoff: Not necessary comment here. 
It could be on `refreshSearchViews` or just removed.

core/search/src/main/kotlin/io/klibs/core/search/configuration/properties/OpenSearchProperties.kt
```diff
@@ -14,4 +14,11 @@ data class OpenSearchProperties(
     val packageIndex: String = "package",
     val username: String? = null,
     val password: String? = null,
+
+    val connectTimeout: Duration = Duration.ofSeconds(5),
+    val socketTimeout: Duration = Duration.ofSeconds(30),
+    val requestTimeout: Duration = Duration.ofSeconds(30),
+
+    val reapMinAge: Duration = Duration.ofHours(1),
```
dkrasnoff: Could you please add a little bit of explanation documentation to the new fields? Old properties are understandable, but some of the new ones are not obvious.

core/search/src/main/kotlin/io/klibs/core/search/configuration/OpenSearchConfiguration.kt
```diff
@@ -21,13 +22,29 @@ import java.net.URI
...
+        ): OpenSearchClient {
         val uri = URI(properties.uri)
         val host = HttpHost(uri.scheme, uri.host, uri.port)
+        val connectTimeout = Timeout.of(properties.connectTimeout)
+        val socketTimeout = Timeout.of(properties.socketTimeout)
+        val requestTimeout = Timeout.of(properties.requestTimeout)
         val transport = ApacheHttpClient5TransportBuilder.builder(host)
-            .setMapper(JacksonJsonpMapper())
+            .setMapper(jacksonJsonpMapper)
+            .setRequestConfigCallback {
```
dkrasnoff: Just in case, why do we need to configure connection timeouts? Why aren't the defaults suitable for us?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexSpec.kt
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Everything one index is built from, plus the names derived from it:
+ *
+ * ```
+ * alias      project-a3f9c1e2
+ * generation project-a3f9c1e2-20260730t101500
+ * ```
+ */
+class IndexSpec(
```
dkrasnoff: nit: let's move to the dto package and rename to `SeachIndexSpec` or `OpenSearchIndexSpec`?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexNaming.kt
```diff
@@ -0,0 +1,27 @@
+package io.klibs.core.search.opensearch
+
+import io.klibs.core.search.configuration.properties.OpenSearchProperties
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+
+/** The indices this version builds, one [IndexSpec] each. */
+@Component
+@ConditionalOnProperty("klibs.search.opensearch.enabled", havingValue = "true")
+class IndexNaming(properties: OpenSearchProperties) {
```
dkrasnoff: I'd suggest removing this class and make `ProjectSearchIndex` and `PackageSearchIndex` as a beans declared in the `OpenSearchConfiguration`. Then instead of using `all` you will be able to just inject this beans as a list directly in the `SearchIndexSync`.
Some parts of the change:
1) In `OpenSearchConfiguration`
```kotlin
   @Bean
    fun projectIndex(properties: OpenSearchProperties) : IndexSpec = IndexSpec(
        base = properties.projectIndex,
        settings = IndexDefinitions.PROJECT_SETTINGS,
        mappings = IndexDefinitions.PROJECT_MAPPINGS,
        sql = IndexDefinitions.PROJECT_DOC_SQL,
    ) { it.get("project_id").asText() }

    @Bean
    fun packageIndexing(properties: OpenSearchProperties) : IndexSpec = IndexSpec(
        base = properties.packageIndex,
        settings = IndexDefinitions.PACKAGE_SETTINGS,
        mappings = IndexDefinitions.PACKAGE_MAPPINGS,
        sql = IndexDefinitions.PACKAGE_DOC_SQL,
    ) { "${it.get("group_id").asText()}:${it.get("artifact_id").asText()}" }
```
In `SearchIndexSync`
```kotlin 
@Component
@ConditionalOnProperty("klibs.search.opensearch.enabled", havingValue = "true")
class SearchIndexSync(
    private val lockingTaskExecutor: LockingTaskExecutor,
    private val indexer: OpenSearchIndexer,
    private val searchIndexSpecs: List<IndexSpec>,
) {

    fun syncAll() = run(searchIndexSpecs)
    ...

```

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexSpec.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    fun generation(now: Instant = Instant.now()): String = "$alias-${TIMESTAMP.format(now)}"
+
+    fun aliasMatches(index: String): Boolean = currentAliasWithTimestamp.matches(index)
+
+    fun timestampOf(index: String): Instant? = runCatching {
+        LocalDateTime.parse(index.takeLast(TIMESTAMP_LENGTH), PARSER).toInstant(ZoneOffset.UTC)
+    }.getOrNull()
+
+    private val currentAliasWithTimestamp = Regex("^${Regex.escape(alias)}-\\d{8}t\\d{6}$")
```
dkrasnoff: nit: not sure that this pattern works. I see that in test we have generation in `2026-07-30T10:00:00Z` format, but this regex expects `2026-07-30T10:00:00`

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    private fun sameBaseIndices(spec: IndexSpec): Map<String, IndexState> =
+        client.indices().get { it.index("${spec.base}*") }.result()
+
+    private fun currentTargets(alias: String): Set<String> {
+        val indicesClient = client.indices()
+        if (!indicesClient.existsAlias { it.name(alias) }.value()) return emptySet()
+
+        val aliasesByIndex = indicesClient.getAlias { it.name(alias) }.result()
+        return aliasesByIndex.keys
```
dkrasnoff, line 93: Should real OpenSearch contain only one element for the given alias? Can we maybe add some check here if yes?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
```diff
@@ -0,0 +1,141 @@
...
+                updateRequest.actions { a -> a.remove { it.index(old).alias(spec.alias) } }
+            }
+            updateRequest.actions { a -> a.add { it.index(generation).alias(spec.alias) } }
+        }
+    }
+
+    private fun reap(spec: IndexSpec, newGen: String, now: Instant, live: String?) {
+        val stale = sameAliasIndices(spec)
+            .filter { it != newGen && it != live }
+            .filter { index -> spec.timestampOf(index)?.isBefore(now.minus(properties.reapMinAge)) == true }
```
dkrasnoff: why do we need this check?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
```diff
@@ -0,0 +1,141 @@
...
+
+        stale.forEach { index ->
+            client.indices().delete { it.index(index) }
+            log.info("reaped stale generation '{}'", index)
+        }
+
+        val staleForeign = sameBaseIndices(spec)
```
dkrasnoff: why don't just combine stale and staleForeign?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
```diff
@@ -0,0 +1,141 @@
...
+    fun sync(spec: IndexSpec, now: Instant = Instant.now()) {
+        val indices = currentTargets(spec.alias)
+        check(indices.size <= 1) {
+            "alias '${spec.alias}' points at ${indices.size} indices $indices; " +
+                    "refusing to rebuild until it points at one"
+        }
+        val current = indices.singleOrNull()
+
+        val generation = spec.generation(now)
+        reap(spec, newGen = generation, now = now, live = current)
```
dkrasnoff: Shouldn't we reap after creating new index?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
```diff
@@ -0,0 +1,141 @@
...
+
+        client.indices().create {
+            it.index(generation)
+                .settings(parse(spec.settings, IndexSettings._DESERIALIZER))
+                .mappings(parse(withMeta(spec.mappings, spec.hash), TypeMapping._DESERIALIZER))
+        }
+
+        val docs = jdbcClient.sql(spec.sql)
```
dkrasnoff: You put all the docs in the memory. Let's make it batch from here

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
```diff
@@ -0,0 +1,141 @@
...
+        check(docs.isNotEmpty()) { "projection for '${spec.alias}' returned no rows; refusing to swap onto an empty index" }
+
+        docs.chunked(BATCH).forEach { batch -> bulk(generation, batch, spec.idOf) }
+        client.indices().refresh { it.index(generation) }
+
+        swapAlias(spec, generation, current)
+        log.info("swapped alias '{}' onto '{}' with {} docs", spec.alias, generation, docs.size)
+    }
+
+    private fun bulk(index: String, batch: List<ObjectNode>, idOf: (ObjectNode) -> String) {
```
dkrasnoff: nit: let's name it bulkIndex()

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        val docs = jdbcClient.sql(spec.sql)
+            .query(String::class.java)
+            .list()
+            .map { mapper.readTree(it) as ObjectNode }
+        check(docs.isNotEmpty()) { "projection for '${spec.alias}' returned no rows; refusing to swap onto an empty index" }
+
+        docs.chunked(BATCH).forEach { batch -> bulk(generation, batch, spec.idOf) }
+        client.indices().refresh { it.index(generation) }
```
dkrasnoff: why do we refresh here, not clear for me, we just put new index to the OpenSearch

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
```diff
@@ -0,0 +1,141 @@
...
+        val response = client.bulk(request.build())
+        check(!response.errors()) {
+            "bulk into '$index' had errors: ${response.items().firstOrNull { it.error() != null }?.error()?.reason()}"
+        }
+    }
+
+    private fun swapAlias(spec: IndexSpec, generation: String, current: String?) {
+        client.indices().updateAliases { updateRequest ->
+            current?.let { old ->
+                updateRequest.actions { a -> a.remove { it.index(old).alias(spec.alias) } }
```
dkrasnoff, line 82: Why does old index have the same alias as new?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexNaming.kt
nikitavlaevjb: This version of the code: mappings, sqls and such

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: Yes and the check is right outside of this method

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: Because we want to keep some of the latest indices to be able to rollback

app/src/main/kotlin/io/klibs/app/search/SearchIndexSync.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    fun buildMissingAliases() = run(naming.all.filterNot { indexer.aliasExists(it) })
+
+    private fun run(targets: List<IndexSpec>) {
+        val failures = targets.mapNotNull { runCatching { withLock(it) }.exceptionOrNull() }
+        failures.forEach { log.error("OpenSearch index sync failed", it) }
+        failures.firstOrNull()?.let { throw it }
+    }
+
+    private fun withLock(spec: IndexSpec) {
```
dkrasnoff: Not clear to me why we need `withLock` logic?

app/src/main/kotlin/io/klibs/app/job/SearchIndexSyncJob.kt
```diff
@@ -0,0 +1,25 @@
...
+class SearchIndexSyncJob(
+    private val searchIndexSync: SearchIndexSync,
+) {
+
+    @Scheduled(initialDelay = 0, fixedRate = 10, timeUnit = TimeUnit.MINUTES)
+    @Timed(
+        value = "klibs.search.opensearch.sync.time",
+        description = "Klibs: Time taken to rebuild the OpenSearch indices and swap their aliases",
+    )
+    fun sync() = searchIndexSync.syncAll()
```
dkrasnoff, line 29: Why don't we use @SchedulerLock just here and add custom locking logic?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: because the index name is like "project-hash-timestamp". "project-hash" is an alias. so we swap alias here between two indices with different timestamps

app/src/main/resources/application.yml
```diff
@@ -29,6 +29,11 @@ spring:
   jpa:
     hibernate:
       ddl-auto: validate
+  task:
+    scheduling:
+      # A single scheduler thread lets long jobs starve the 10-minute search sync indefinitely.
```
dkrasnoff: And what happens when it is not a single scheduler thread ?

app/src/main/kotlin/io/klibs/app/search/SearchIndexBootstrap.kt
```diff
@@ -0,0 +1,40 @@
...
+        searchIndexReadiness.refresh()
+    }
+
+    override fun stop() {
+        running = false
+    }
+
+    override fun isRunning(): Boolean = running
+
+    /** Below the web server's own phase (`DEFAULT_PHASE - 2048`), so it starts first. */
```
dkrasnoff: It looks like a magic number based on some internal implementation `DEFAULT_PHASE - 2048`. What if just make this `SearchIndexBootstrap` as a component or job and depend on something, that we need to run it?

app/src/main/kotlin/io/klibs/app/search/SearchIndexBootstrap.kt
```diff
@@ -0,0 +1,40 @@
+package io.klibs.app.search
```
dkrasnoff, line 1: can all of the classes from this package live in a dedicated module?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: because for foreign indices we need to additionally check if they are in use or not.
Also the reapAge is different

core/search/src/main/resources/opensearch/project-doc.sql
```diff
@@ -0,0 +1,72 @@
...
+),
+tags_info AS (
+  SELECT project_id, COALESCE(
+           array_agg(DISTINCT value ORDER BY value DESC) FILTER (WHERE origin = 'USER'),
+           array_agg(DISTINCT value ORDER BY value DESC) FILTER (WHERE origin = 'GITHUB'),
+           array_agg(DISTINCT value ORDER BY value DESC) FILTER (WHERE origin = 'AI')
+         ) AS tags
+  FROM project_tags GROUP BY project_id
+)
+SELECT json_build_object(
```
dkrasnoff, line 41: It should become a project and package index, right? 
Why don't we just use mat view then?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: That is debatable actually.
I wanted to discuss it.
For now, I decided like that because if the index build throws, we have partially built index to inspect.
If we reap in `finally` block, that partial build would be gone.

But for the most part, it does not really matter to reap before or after.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexSpec.kt
nikitavlaevjb: No, the first string that you've provided is an Instant.
The regex is dealing with something that is already formatted (see [io.klibs.core.search.opensearch.IndexSpec#PARSER](https://github.com/JetBrains/klibs-io/blob/fad6c2f27076f5bfcf3b336b73178cf722d823a4/core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexSpec.kt#L44))

core/search/src/main/resources/opensearch/project-doc.sql
nikitavlaevjb: Because matview
1. is suited for the PG fts search
2. (I wish) will be deleted after the replicas appear

Also I think that depending on the matview (especially here) is an antipattern

app/src/main/kotlin/io/klibs/app/search/SearchIndexBootstrap.kt
nikitavlaevjb: Yes it does, and I kinda hate this implementation too.
I would gladly do something else here, you have any ideas?

core/search/src/main/kotlin/io/klibs/core/search/configuration/OpenSearchConfiguration.kt
nikitavlaevjb: Actually, thanks for questioning!
The original point is that the Apache HC5 default were way too loose for our case, but turns out that opensearch-java provides its own ConnectionConfig with good defaults.
So yeah, I'll get rid of those changes

app/src/main/kotlin/io/klibs/app/job/SearchIndexSyncJob.kt
dkrasnoff, line 29: Please, add a log at least debug level for start and finish.
Also, it make sense to measure the time of this job.

nikitavlaevjb: The lock's name is decided in runtime: `searchIndexSync-<base>-<hash>`, afaiu @SchedulerLock cannot express that

app/src/main/kotlin/io/klibs/app/search/SearchIndexSync.kt
nikitavlaevjb: Only one pod should handle the sync at the time

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: After indexing the documents, they still live in the write buffer until refresh and are not yet searchable.
Refresh usually happens often (every 1s), but the thing is that in the next line we instantly `swapAlias`.
That might cause some problems and readers could see empty index for a bit.
So I'd leave it.

app/src/main/resources/application.yml
nikitavlaevjb: Good catch, lets dicuss it

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexSpec.kt
dkrasnoff: val TIMESTAMP_FORMATTER: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMdd't'HHmmss").withZone(ZoneOffset.UTC)

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexSpec.kt
```diff
@@ -0,0 +1,47 @@
...
+    val settings: String,
+    val mappings: String,
+    val sql: String,
+    val idOf: (ObjectNode) -> String,
+) {
+
+    /** Covers everything that shapes a doc, so a contract change lands on a fresh alias. */
+    val hash = IndexDefinitions.shortSha256Hex(settings, mappings, sql)
+
+    val alias = "$base-$hash"
```
dkrasnoff: Could you add doc about difference and reason for alias and generation in our system

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexNaming.kt
dkrasnoff: remove

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
dkrasnoff: currentTargets -> currentGenerations

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
dkrasnoff: Up to you remove or leave this logic

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
dkrasnoff: `staleGenaration` and `staleAlias`

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
dkrasnoff: Discussed that Nikita will decide what to do with it: add comment or just remove this cool logic

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
dkrasnoff: Discussed in DM. Add a comment on why, please.

app/src/main/kotlin/io/klibs/app/configuration/SchedulingConfiguration.kt
```diff
@@ -14,6 +16,11 @@ import javax.sql.DataSource
 @Configuration
 class SchedulingConfiguration {
 
+    /** For locks whose name is only known at runtime, which `@SchedulerLock` cannot express. */
```
dkrasnoff: maybe update this doc and mention the exact situation when we need this

app/src/main/kotlin/io/klibs/app/search/SearchIndexBootstrap.kt
dkrasnoff: Try to create a bean / component and set @Order

core/search/src/main/resources/opensearch/project-doc.sql
dkrasnoff: What if we change column name or other data in PG

app/src/main/kotlin/io/klibs/app/job/SearchIndexSyncJob.kt
dkrasnoff: give me link to next issue

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexNaming.kt
nikitavlaevjb: done

app/src/main/kotlin/io/klibs/app/controller/PingController.kt
nikitavlaevjb: done

core/search/src/main/kotlin/io/klibs/core/search/service/SearchService.kt
nikitavlaevjb: removed

core/search/src/main/kotlin/io/klibs/core/search/configuration/properties/OpenSearchProperties.kt
nikitavlaevjb: done

core/search/src/main/kotlin/io/klibs/core/search/configuration/OpenSearchConfiguration.kt
nikitavlaevjb: removed unneccesary; kept requestTimeout, because default 3min is too long.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexSpec.kt
nikitavlaevjb: done

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexNaming.kt
nikitavlaevjb: done

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexSpec.kt
nikitavlaevjb: done

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: done

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: So I simplified it a bit, now instead of timestamps it has 2 indices that it swaps between

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: commented

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: I tried to add actual streams: too much code and complexity for 9mb of strings.
Optimized parsing a bit, now we store actually at most 9mb instead of 50mb.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: done

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: added

app/src/main/kotlin/io/klibs/app/search/SearchIndexSync.kt
nikitavlaevjb: What did I need to change here, sorry?

app/src/main/resources/application.yml
nikitavlaevjb: Added a separate scheduler

app/src/main/kotlin/io/klibs/app/search/SearchIndexBootstrap.kt
nikitavlaevjb: Done, no need to add @order though

core/search/src/main/resources/opensearch/project-doc.sql
nikitavlaevjb: There are unit test that will catch it

app/src/main/kotlin/io/klibs/app/job/SearchIndexSyncJob.kt
nikitavlaevjb: [KTL-4712](https://youtrack.jetbrains.com/issue/KTL-4712) Search OpenSearch observability

core/search/src/main/kotlin/io/klibs/core/search/opensearch/IndexSpec.kt
nikitavlaevjb: done

app/src/main/kotlin/io/klibs/app/configuration/SchedulingConfiguration.kt
nikitavlaevjb: dobe

app/src/main/kotlin/io/klibs/app/search/SearchIndexBootstrap.kt
nikitavlaevjb: I would say that those classes wire search to app, they don't contain search-specific logic.
I'd keep them as is

core/search/src/main/kotlin/io/klibs/core/search/repository/PackageSearchRepositoryOpenSearch.kt
```diff
@@ -24,10 +25,11 @@ import java.time.ZoneOffset
 @ConditionalOnProperty("klibs.search.opensearch.enabled", havingValue = "true")
 class PackageSearchRepositoryOpenSearch(
     client: OpenSearchClient,
-    properties: OpenSearchProperties,
+    @Qualifier("packageIndexSpec")
```
dkrasnoff: nit: `packageIndexSpec: OpenSearchIndexSpec` will allow you not to use `@Qualifier("packageIndexSpec")`, the same about the project

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
dkrasnoff: Why do we need batching logic further then?

app/src/main/kotlin/io/klibs/app/search/SearchIndexBootstrap.kt
dkrasnoff: > I would say that those classes are wired to app

Yeah, here is a trick. 
I'd like to make the search module easily injectable in any project, so that you just need to add the dependency to any project, like an app, and that's all. 
Or do we have any technical problems with moving it as well?

dkrasnoff (changes requested):

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
nikitavlaevjb: If I understand your question correctly, opensearch recommends using 5-15mb per bulk requests
https://docs.opensearch.org/latest/api-reference/document-apis/bulk/#optimal-batch-size

app/src/main/kotlin/io/klibs/app/search/SearchIndexBootstrap.kt
nikitavlaevjb: You quoted me wrong :)
The main problem is that you need to run jobs to refresh search data. For that reason you need schedlock dependency and a table created for it, scheduling -- that makes that part of search already not independent, because it is **wiring** search logic to app.

nikitavlaevjb: Still, added a refactoring, check the fixup commit

core/search/src/main/kotlin/io/klibs/core/search/opensearch/OpenSearchIndexer.kt
dkrasnoff: Yeah, you understood correctly, thanks!

dkrasnoff (approved):

PR #410, opened 2026-08-19
KTL-4712 opensearch observability: app-part

nikitavlaevjb:
Infra part is coming to the infra repo

### DIFF
--- a/core/search/src/main/kotlin/io/klibs/core/search/opensearch/SearchIndexSync.kt
+++ b/core/search/src/main/kotlin/io/klibs/core/search/opensearch/SearchIndexSync.kt
@@ -1,6 +1,7 @@
 package io.klibs.core.search.opensearch
 
 import io.klibs.core.search.dto.opensearch.OpenSearchIndexSpec
+import io.klibs.core.search.opensearch.metrics.SearchIndexMetrics
 import org.slf4j.LoggerFactory
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.stereotype.Component
@@ -13,6 +14,7 @@ class SearchIndexSync(
     private val searchIndexLock: SearchIndexLock,
     private val indexer: OpenSearchIndexer,
     private val indexSpecs: List<OpenSearchIndexSpec>,
+    private val metrics: SearchIndexMetrics,
 ) {
 
     fun syncAll() = run(indexSpecs)
@@ -20,7 +22,9 @@ class SearchIndexSync(
     fun buildMissingAliases() = run(indexSpecs.filterNot { indexer.aliasExists(it) })
 
     private fun run(targets: List<OpenSearchIndexSpec>) {
-        val failures = targets.mapNotNull { runCatching { withLock(it) }.exceptionOrNull() }
+        val failures = targets.mapNotNull { spec ->
+            runCatching { withLock(spec) }.exceptionOrNull()?.also { metrics.recordFailure(spec) }
+        }
         failures.forEach { log.error("OpenSearch index sync failed", it) }
         failures.firstOrNull()?.let { throw it }
     }
@@ -28,10 +32,16 @@ class SearchIndexSync(
     private fun withLock(spec: OpenSearchIndexSpec) {
         val lock = LockSpec("searchIndexSync-${spec.base}-${spec.hash}", LOCK_AT_MOST_FOR, LOCK_AT_LEAST_FOR)
         val startedAt = Instant.now()
-        if (!searchIndexLock.runLocked(lock) { indexer.sync(spec) }) {
+        var rebuildDuration: Duration? = null
+        if (!searchIndexLock.runLocked(lock) {
+            val syncStartedAt = Instant.now()
+            indexer.sync(spec)
+            rebuildDuration = Duration.between(syncStartedAt, Instant.now())
+        }) {
             log.info("another pod holds '{}', skipping this run", lock.name)
             return
         }
+        rebuildDuration?.let { metrics.recordSuccess(spec, it) }
 
         val elapsed = Duration.between(startedAt, Instant.now())
         if (elapsed > LOCK_AT_MOST_FOR) {
--- a/core/search/src/main/kotlin/io/klibs/core/search/opensearch/metrics/SearchIndexMetrics.kt
+++ b/core/search/src/main/kotlin/io/klibs/core/search/opensearch/metrics/SearchIndexMetrics.kt
@@ -1,0 +1,76 @@
+package io.klibs.core.search.opensearch.metrics
+
+import io.klibs.core.search.dto.opensearch.OpenSearchIndexSpec
+import io.klibs.core.search.opensearch.OpenSearchIndexer
+import io.micrometer.core.instrument.Counter
+import io.micrometer.core.instrument.Gauge
+import io.micrometer.core.instrument.MeterRegistry
+import io.micrometer.core.instrument.Tags
+import io.micrometer.core.instrument.Timer
+import org.slf4j.LoggerFactory
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+import java.time.Duration
+import java.time.Instant
+
+/**
+ * Freshness and outcome of the OpenSearch index rebuild.
+ */
+@Component
+@ConditionalOnProperty("klibs.search.opensearch.enabled", havingValue = "true")
+class SearchIndexMetrics(
+    private val registry: MeterRegistry,
+    private val indexer: OpenSearchIndexer,
+    indexSpecs: List<OpenSearchIndexSpec>,
+) {
+
+    init {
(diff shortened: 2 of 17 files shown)


core/search/src/main/kotlin/io/klibs/core/search/opensearch/SearchIndexMetrics.kt
```diff
@@ -0,0 +1,77 @@
...
+        Timer.builder(SYNC_DURATION)
+            .description("Klibs: Time to rebuild one index and swap its alias")
+            .tags(tagsOf(spec))
+            .register(registry)
+            .record(took)
+    }
+
+    fun recordFailure(spec: OpenSearchIndexSpec) {
+        Counter.builder(SYNC_FAILURES)
+            .description("Klibs: Index rebuilds that failed")
```
dkrasnoff: nit: "Number of failed open search's index rebuilds"

core/search/src/main/kotlin/io/klibs/core/search/opensearch/SearchIndexSync.kt
```diff
@@ -13,25 +13,34 @@ class SearchIndexSync(
...
+        }
         failures.forEach { log.error("OpenSearch index sync failed", it) }
         failures.firstOrNull()?.let { throw it }
     }
 
     private fun withLock(spec: OpenSearchIndexSpec) {
         val lock = LockSpec("searchIndexSync-${spec.base}-${spec.hash}", LOCK_AT_MOST_FOR, LOCK_AT_LEAST_FOR)
         val startedAt = Instant.now()
-        if (!searchIndexLock.runLocked(lock) { indexer.sync(spec) }) {
+        var rebuild: Pair<Int, Duration>? = null
```
dkrasnoff: nit: rebuild -> rebuildData or rebuildResult or rebuildInfo

core/search/src/main/kotlin/io/klibs/core/search/opensearch/SearchIndexMetrics.kt
```diff
@@ -0,0 +1,77 @@
...
+            .tags(tagsOf(spec))
+            .register(registry)
+            .increment()
+    }
+
+    private fun stateOf(spec: OpenSearchIndexSpec): IndexState = states.computeIfAbsent(spec.alias) {
+        IndexState().also { state ->
+            Gauge.builder(INDEX_AGE, state) { it.ageSeconds() }
```
dkrasnoff: Do we really want to collect the INDEX_AGE metric this way? I mean we have 3 pods, and each of them can rebuild the index.
Are there other ways to collect this metric? For example, during the indices swap to check the creation date of the old index?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/SearchQueryMetrics.kt
```diff
@@ -0,0 +1,73 @@
+package io.klibs.core.search.opensearch
```
dkrasnoff: Lets move metrics to a dedicated package? I'm starting to get lost in the classes of this package.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/metrics/SearchQueryMetrics.kt
```diff
@@ -0,0 +1,73 @@
...
+            throw e
+        } finally {
+            val took = Duration.ofNanos(System.nanoTime() - startedAt)
+            Timer.builder(QUERY_TIME)
+                .description("Klibs: OpenSearch query round trip")
+                .tags(tagsOf(spec))
+                .register(registry)
+                .record(took)
+            if (took > slowQueryThreshold) {
+                log.warn(
```
dkrasnoff, line 45: just in case: will we have any alert based on this warning message? Or why QUERY_TIME metric will not be enough?

dkrasnoff:
Good job, it seems that you covered all the necessary places 👍 👍 👍

dkrasnoff (changes requested):

core/search/src/main/kotlin/io/klibs/core/search/opensearch/metrics/SearchQueryMetrics.kt
nikitavlaevjb: This log tells us which queries were slow exactly.
The alert will be on klibs.search.query.time

core/search/src/main/kotlin/io/klibs/core/search/opensearch/SearchIndexMetrics.kt
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/opensearch/SearchIndexSync.kt
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/opensearch/SearchQueryMetrics.kt
nikitavlaevjb: fixed

core/search/src/main/kotlin/io/klibs/core/search/opensearch/SearchIndexMetrics.kt
nikitavlaevjb: Good catch!
In reality, they were merged on the alert level under min(...), which eliminates nulls.
Still, that was a poor design. Fixed by quering actual index life length.

core/search/src/main/kotlin/io/klibs/core/search/opensearch/metrics/SearchIndexMetrics.kt
```diff
@@ -21,14 +22,25 @@ import java.util.concurrent.atomic.AtomicReference
...
+                .tags(tagsOf(spec))
+                .register(registry)
+        }
+    }
 
     fun recordSuccess(spec: OpenSearchIndexSpec, docCount: Int, took: Duration) {
-        val state = stateOf(spec)
-        state.docs.set(docCount.toLong())
-        state.lastSuccess.set(Instant.now())
+        states.getValue(spec.alias).docs.set(docCount.toLong())
```
dkrasnoff: Why do we get the value instead of calculating it here?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/metrics/SearchIndexMetrics.kt
```diff
@@ -44,29 +56,33 @@ class SearchIndexMetrics(
...
+                .getOrNull()
+            cachedCreatedAt.set(now to createdAt)
+            return createdAt
+        }
     }
 
     private companion object {
+        private val log = LoggerFactory.getLogger(SearchIndexMetrics::class.java)
+
+        private val CACHE_TTL: Duration = Duration.ofSeconds(30)
```
dkrasnoff: Not clear for me, why do we need `CACHE_TTL`?

core/search/src/main/kotlin/io/klibs/core/search/opensearch/metrics/SearchIndexMetrics.kt
```diff
@@ -44,29 +56,33 @@ class SearchIndexMetrics(
...
-                .tags(tagsOf(spec))
-                .register(registry)
-        }
-    }
-
-    private class IndexState {
-        val lastSuccess = AtomicReference<Instant>()
+    private inner class IndexState(private val spec: OpenSearchIndexSpec) {
         val docs = AtomicLong()
+        private val cachedCreatedAt = AtomicReference<Pair<Instant, Instant?>>()
```
dkrasnoff: Why do we need to have AtomicReference, if we calculate `ageSeconds` only once and get the.
The whole this logic seem to be strange to me, why don't just leave:
```
        fun ageSeconds(): Double = createdAt()
            ?.let { Duration.between(it, Instant.now()).seconds.toDouble() }
            ?: Double.NaN

        private fun createdAt(): Instant? {
            val createdAt = runCatching { indexer.servingIndexCreatedAt(spec) }
                .onFailure { log.debug("could not read the creation date of '{}'", spec.alias, it) }
                .getOrNull()
            return createdAt
        }
```

core/search/src/main/kotlin/io/klibs/core/search/opensearch/metrics/SearchIndexMetrics.kt
dkrasnoff: I mean, you know `oldIndex` name (it is in the `io.klibs.core.search.opensearch.OpenSearchIndexer#sync`, let's just use it.

dkrasnoff (changes requested):

dkrasnoff (approved):

PR #427, opened 2026-08-26
Added retry after header processing in Maven Central rate limiter.

dkrasnoff:


integrations/maven/src/main/kotlin/io/klibs/integration/maven/request/impl/MavenCentralRateLimiter.kt
```diff
@@ -2,24 +2,31 @@ package io.klibs.integration.maven.request.impl
...
+import kotlin.time.Clock
+import kotlin.time.Duration.Companion.seconds
+import kotlin.time.toJavaDuration
 
 @Component
 class MavenCentralRateLimiter(
     private val mavenIntegrationProperties: MavenIntegrationProperties,
-    meterRegistry: MeterRegistry
+    meterRegistry: MeterRegistry,
+    private val clock: Clock = Clock.System,
```
nikitavlaevjb, line 20: Why Clock, not Instant?

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClient.kt
```diff
@@ -196,14 +201,22 @@ abstract class BaseMavenSearchClient(
...
 
                     else -> throw IOException("Unexpected response: ${response.code}")
                 }
             }
         }
     }
 
+    private fun applyRequestCooldown(response: Transport.Response, serviceUri: String): Nothing {
+        val retryAfter = parseRetryAfter(response)
+        retryAfter?.let { rateLimiter.applyCooldown(it) }
```
nikitavlaevjb, line 217: Add logging here

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClient.kt
```diff
@@ -196,14 +201,22 @@ abstract class BaseMavenSearchClient(
                         followRedirects(location, headers, converter, redirectCount + 1, requestExecutor)
                     }
 
-                    HTTP_TOO_MANY_REQUESTS -> throw MavenRateLimitedException(serviceUri)
+                    HTTP_TOO_MANY_REQUESTS -> {
+                        applyRequestCooldown(response, serviceUri)
```
nikitavlaevjb, line 205: We have 3 CDNs actually under one ratelimiter 
- repo1.maven.org: Cloudflare,
- search.maven.org: Cloudfront,
- central.sonatype.com: smth else.
If we share the MavenCentralRateLimiter, maybe they all should be able to applyRequestCooldown?

nikitavlaevjb (changes requested):

integrations/maven/src/main/kotlin/io/klibs/integration/maven/request/impl/MavenCentralRateLimiter.kt
dkrasnoff: Because Instant.now() is deprecated

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClient.kt
dkrasnoff: I expect that we never use:
```
search.maven.org: Cloudfront,
central.sonatype.com: smth else.
```
, so the only actual place now is `repo1.maven.org`

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClient.kt
dkrasnoff: added

nikitavlaevjb (approved):

PR #433, opened 2026-08-28
Added Google mirror for Maven Central

dkrasnoff:
Also two first commits in this PR includes just removing the dead code


app/src/main/kotlin/io/klibs/app/service/impl/CentralSonatypeUserIndexingRequestService.kt
```diff
@@ -0,0 +1,33 @@
...
+import io.klibs.integration.maven.search.impl.BaseCentralMavenSearchClient
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Service
+
+@Service
+@ConditionalOnProperty(
+    name = ["klibs.indexing-configuration.google-maven-central-mirror.enabled"],
+    havingValue = "false",
+    matchIfMissing = true
```
nikitavlaevjb: There are 2 UserIndexingRequestService beans now: CentralSonatypeUserIndexingRequestService and GoogleMavenCentralMirrorUserIndexingRequestService.
What if they are turned on together?
Is that a conflict?

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientRedirectTest.kt
```diff
@@ -147,6 +149,49 @@ class BaseMavenSearchClientRedirectTest {
...
+        val result = client.getPomWithReleaseDate(
+            MavenArtifact("org.example", "example-artifact", "1.0.0", ScraperType.CENTRAL_SONATYPE)
+        )
+
+        assertNotNull(result)
+        assertEquals(Instant.parse("2021-02-13T16:31:54Z").toJavaInstant(), result.releasedAt)
+    }
+
+    @Test
+    fun `release date prefers google mirror metadata over storage last-modified`() {
```
nikitavlaevjb: This test tests behavior that only exists in the test.
GoogleMirrorTestClient.getReleasedAt does not align with whats happening in the actual GoogleMavenCentralMirrorSearchClient, right?

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/GoogleMavenCentralMirrorSearchClient.kt
```diff
@@ -0,0 +1,29 @@
...
+    @Value("\${klibs.integration.maven.central.content-fallback-endpoint}")
+    contentFallbackEndpoint: String,
+) : BaseCentralMavenSearchClient(
+    xmlMapper,
+    mavenCentralRateLimiter,
+    LoggerFactory.getLogger(GoogleMavenCentralMirrorSearchClient::class.java),
+    objectMapper,
+    contentEndpoint,
+    contentFallbackEndpoint,
+    "x-goog-meta-last-modified"
```
nikitavlaevjb: This single header string is the entire difference from `CentralSonatypeSearchClient`.
Looks like we dont need those beans separated

Also, unused imports in this file.

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/GoogleMavenCentralMirrorSearchClient.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlin.time.Instant
+import org.apache.maven.search.api.transport.Transport
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+
+@Component("GOOGLE_MAVEN_CENTRAL_MIRROR")
+class GoogleMavenCentralMirrorSearchClient(
+    xmlMapper: XmlMapper,
+    mavenCentralRateLimiter: MavenCentralRateLimiter,
```
nikitavlaevjb, line 18: Dont reuse `MavenCentralRateLimiter` here? measure some kind of separate rate limit?

app/src/main/resources/application.yml
```diff
@@ -120,6 +119,8 @@ klibs:
       regen-ttl: ${KLIBS_DESCRIPTION_REGEN_TTL:90d}
     central-sonatype:
       enabled: ${KLIBS_ENABLE_INDEXING_CENTRAL_SONATYPE:false}
+    google-maven-central-mirror:
```
nikitavlaevjb: So google-maven-central-mirror and regular maven central downloader can be enable together and download the same index in theory?
Would that break something? looks weird

app/src/main/kotlin/io/klibs/app/service/impl/BaseUserIndexingRequestService.kt
```diff
@@ -38,9 +37,9 @@ internal class DefaultUserIndexingRequestService(
         fulfillRequest(userRequestIssue.groupId, userRequestIssue.artifactId, userRequestIssue.version, userRequestIssue)
     }
 
-    internal fun fulfillRequest(
+    private fun fulfillRequest(
         groupId: String,
-        artifactId: String?,
+        artifactId: String,
```
nikitavlaevjb: Removed nullability created dead code

app/src/main/kotlin/io/klibs/app/service/impl/DefaultUserIndexingRequestService.kt
```diff
@@ -11,23 +11,22 @@ import io.klibs.core.project.blacklist.BlacklistRepository
...
 import org.slf4j.LoggerFactory
-import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
 import java.util.UUID
 import kotlin.jvm.optionals.getOrNull
 
-@Service
-internal class DefaultUserIndexingRequestService(
-    private val centralSonatypeSearchClient: CentralSonatypeSearchClient,
+internal abstract class DefaultUserIndexingRequestService(
```
nikitavlaevjb: This is BaseUserIndexingRequestService now

app/src/main/kotlin/io/klibs/app/indexing/discoverer/impl/GoogleMavenMavenCentralMirrorPackageDiscoverer.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+
+@Component
+@ConditionalOnProperty(
+    name = ["klibs.indexing-configuration.google-maven-central-mirror.enabled"],
+    havingValue = "true",
+    matchIfMissing = false
+)
+class GoogleMavenMavenCentralMirrorPackageDiscoverer(
```
nikitavlaevjb: 1. Class name has "Maven" twice
2. Beans are resolved by parameter name (`googleMavenCentralMirrorIndexScannerService`, `googleMavenCentralMirrorScraper`). 
This works today  because the `@ConditionalOnProperty`s leave a single candidate of each type
3. `MavenCentralSonatypePackageDiscoverer` was renamed but its test file is still `CentralSonatypePackageDiscovererTest`.

app/src/main/resources/application-local.yml
```diff
@@ -35,8 +35,10 @@ klibs:
       thread-count: 2
     central-sonatype:
       enabled: false
+    google-maven-central-mirror:
+      enabled: false
     gmaven:
-      enabled: true
+      enabled: false
```
nikitavlaevjb, line 42: accidental?

app/src/main/kotlin/io/klibs/app/indexing/discoverer/impl/BaseMavenCentralPackageDiscoverer.kt
```diff
@@ -97,11 +91,9 @@ class CentralSonatypePackageDiscoverer(
         }
     }
 
-
     private fun MavenArtifact.getArtifactCoordinates(): String = createArtifactCoordinates(groupId, artifactId)
 
-
     private companion object {
-        private val logger = LoggerFactory.getLogger(CentralSonatypePackageDiscoverer::class.java)
+        private val logger = LoggerFactory.getLogger(BaseMavenCentralPackageDiscoverer::class.java)
```
nikitavlaevjb: The logger is bound to the base class, so both sources now log under `BaseMavenCentralPackageDiscoverer`. The `$sourceName` prefix in the messages covers most of it, but `LoggerFactory.getLogger(this::class.java)` (non-companion) would keep per-source logger names and make log filtering work.

integrations/maven/src/main/kotlin/io/klibs/integration/maven/scraper/impl/BaseMavenCentralScraper.kt
```diff
@@ -0,0 +1,61 @@
...
+            )
+            if (exception is MavenRateLimitedException) {
+                logger.warn("Rate limited by Maven Central, stopping new version discovery")
+                break
+            }
+        }
+    }
+
+    protected companion object {
+        val logger = LoggerFactory.getLogger(BaseMavenCentralScraper::class.java)
```
nikitavlaevjb: why companion object is protected and logger is public?
Something weird with modifiers here

app/src/test/kotlin/io/klibs/app/service/impl/UserIndexingRequestServiceTest.kt
```diff
@@ -217,95 +200,64 @@ class UserIndexingRequestServiceTest : BaseUnitWithDbLayerTest() {
...
 
         val exception = assertThrows<UserRequestProcessingException> {
-            uut.fulfillRequest("com.banned", null, null)
+            fulfillRequest(groupId = "com.banned", artifactId = "libX", version = null)
         }
 
         assertEquals("All artifacts from this request are already indexed, queued or banned", exception.reason)
     }
+
+    private fun fulfillRequest(
```
nikitavlaevjb, line 265: "some banned, some saved" path is not covered anymore here.
Now it is the only the all-banned path.

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/BaseCentralMavenSearchClient.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.slf4j.Logger
+
+abstract class BaseCentralMavenSearchClient(
+    xmlMapper: XmlMapper,
+    mavenCentralRateLimiter: MavenCentralRateLimiter,
+    logger: Logger,
+    objectMapper: ObjectMapper,
+    private val contentEndpoint: String,
+    private val contentFallbackEndpoint: String,
+    lastModifiedHeader: String,
```
nikitavlaevjb: Could you clarify the intent behind a class per source here?

As it stands, the whole delta between `CentralSonatypeSearchClient` and `GoogleMavenCentralMirrorSearchClient` is this one string. Same `content-endpoint` property, same `content-fallback-endpoint`, same `MavenCentralRateLimiter`, same `SmoSearchBackend` pointed at `central.sonatype.com`. The only thing the subclasses buy is a bean name for `@Qualifier` and a logger name -- both of which two `@Bean` factory methods in a `@Configuration` would give you, with the per-source config visible in one place instead of spread across `@Value`s in two classes.

But the test's `GoogleMirrorTestClient` suggests the mirror is meant to have genuinely different `getReleasedAt` logic (epoch header + fallback chain), which would justify a subclass and make the `lastModifiedHeader` parameter redundant.

So the two shapes seem to be:

- **A** -- the mirror really does need different behavior: move the `getReleasedAt` override from the test into `GoogleMavenCentralMirrorSearchClient`, drop `lastModifiedHeader` here (or keep it only as the base's default-impl input), keep both classes.
- **B** -- the mirror is pure config: drop both subclasses, one `CentralMavenSearchClient` + two `@Bean`s, with per-source endpoint and header in properties (which would also address the shared-endpoint point above).

My read is A, since the GCS metadata handling is genuinely different -- but right now the PR is in between: a class per source *and* the behavior it exists for sitting in a test.

The same question applies to `CentralSonatypeMavenIndexScannerService` / `GoogleMavenCentralMirrorIndexScannerService` and the two scrapers, where the delta is one enum constant.

integrations/maven/src/main/kotlin/io/klibs/integration/maven/service/CentralSonatypeMavenIndexScannerService.kt
```diff
@@ -0,0 +1,17 @@
...
+@Service
+@ConditionalOnProperty(
+    name = ["klibs.indexing-configuration.central-sonatype.enabled"],
+    havingValue = "true",
+    matchIfMissing = false
+)
+class CentralSonatypeMavenIndexScannerService(
+    indexer: Indexer,
+    indexingContextManager: MavenIndexingContextManager,
+) : MavenIndexScannerService(indexer, indexingContextManager, ScraperType.CENTRAL_SONATYPE)
```
nikitavlaevjb: This class (and `GoogleMavenCentralMirrorIndexScannerService`, `CentralSonatypeScraper`, `GoogleMavenCentralMirrorScraper`) exists only to supply one enum constant to the base constructor. Four classes for four constants.

`@Bean` factory methods in a config class would express the same thing without the type hierarchy -- and would force the per-source `index-dir` / `index-endpoint` question (see my comment on `application.yml`) into one visible place.

integrations/maven/src/main/kotlin/io/klibs/integration/maven/scraper/impl/CentralSonatypeScraper.kt
```diff
@@ -23,89 +12,5 @@ import java.time.Instant
...
-        }
-    }
-
-    companion object {
-        private val logger = org.slf4j.LoggerFactory.getLogger(CentralSonatypeScraper::class.java)
-    }
-}
\ No newline at end of file
+    centralSonatypeSearchClient: CentralSonatypeSearchClient,
+) : BaseMavenCentralScraper(centralSonatypeSearchClient, ScraperType.CENTRAL_SONATYPE)
```
nikitavlaevjb: Nit: missing newline at EOF. Same in `MavenCentralSonatypePackageDiscoverer.kt`, `CentralSonatypeSearchClient.kt`, `GoogleMavenCentralMirrorSearchClient.kt`, `BaseMavenSearchClient.kt` and `UserIndexingRequestServiceTest.kt`.

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/GoogleMavenCentralMirrorSearchClient.kt
nikitavlaevjb: Maybe use @Bean factory methods in a config class?

nikitavlaevjb (approved):

app/src/main/kotlin/io/klibs/app/service/impl/CentralSonatypeUserIndexingRequestService.kt
dkrasnoff: The Idea, is that they shouldn't be turned on together, because they are the same one.
And you will not make a mistake, because application will just not start if the are both enabled:)

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientRedirectTest.kt
dkrasnoff: You are totally right, thanks!

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/GoogleMavenCentralMirrorSearchClient.kt
dkrasnoff: If I understand your question correctly, then the answer is: there is no sense in using a separate ratelimiter, the current one is fully configurable via settings, and semantically created for Maven Central scraping (google in this case could be considered as real Maven Central)

app/src/main/resources/application.yml
dkrasnoff: They can not be enabled together, as Spring will not be able to inject them both during the application start, so the application will not start and we will see it.

app/src/main/kotlin/io/klibs/app/service/impl/BaseUserIndexingRequestService.kt
dkrasnoff: Could you elaborate on which code, please? I don't see :(

app/src/main/resources/application-local.yml
dkrasnoff: No, just don't need to always have it locally enabled (creates unnecessary logs for me)

app/src/test/kotlin/io/klibs/app/service/impl/UserIndexingRequestServiceTest.kt
dkrasnoff: Could you elaborate about which exact test you are talking here

app/src/main/kotlin/io/klibs/app/service/impl/CentralSonatypeUserIndexingRequestService.kt
nikitavlaevjb: Application started ok for me with both flags enabled

app/src/main/kotlin/io/klibs/app/service/impl/BaseUserIndexingRequestService.kt
nikitavlaevjb: make artifactId not-null in discoverArtifacts and you will see

app/src/test/kotlin/io/klibs/app/service/impl/UserIndexingRequestServiceTest.kt
nikitavlaevjb: Sorry, I didnt notice that corresponding functionality was removed also

app/src/main/kotlin/io/klibs/app/service/impl/CentralSonatypeUserIndexingRequestService.kt
dkrasnoff: Then I've made a mistake, will check, thanks

app/src/main/kotlin/io/klibs/app/service/impl/BaseUserIndexingRequestService.kt
dkrasnoff: Removed, still it is used

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/GoogleMavenCentralMirrorSearchClient.kt
dkrasnoff: I expect, that bean factories shouldn't be used [without real necessity](https://docs.spring.io/spring-framework/reference/core/beans/beanfactory.html#context-introduction-ctx-vs-beanfactory).

app/src/main/kotlin/io/klibs/app/service/impl/CentralSonatypeUserIndexingRequestService.kt
dkrasnoff: Fixed, now it should fail, when both enabled/disabled, in all other cases should work.

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientRedirectTest.kt
dkrasnoff: fixed

app/src/main/resources/application.yml
dkrasnoff: fixed to make them not be enabled together

app/src/main/kotlin/io/klibs/app/service/impl/BaseUserIndexingRequestService.kt
dkrasnoff: a little bit updated this code, but still no dead code

app/src/main/kotlin/io/klibs/app/service/impl/DefaultUserIndexingRequestService.kt
dkrasnoff: fixed

app/src/main/kotlin/io/klibs/app/indexing/discoverer/impl/GoogleMavenMavenCentralMirrorPackageDiscoverer.kt
dkrasnoff: fixed

integrations/maven/src/main/kotlin/io/klibs/integration/maven/scraper/impl/BaseMavenCentralScraper.kt
dkrasnoff: fixed

app/src/main/kotlin/io/klibs/app/service/impl/CentralSonatypeUserIndexingRequestService.kt
nikitavlaevjb: set one to true in application-local.yml then

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/MavenSearchClient.kt
```diff
@@ -9,19 +9,4 @@ import java.time.Instant
 interface MavenSearchClient
 {
     fun pageSize(): Int
```
nikitavlaevjb: pageSize() is not used too, right?

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClient.kt
```diff
@@ -266,23 +267,32 @@ abstract class BaseMavenSearchClient(
...
+        return try {
+            parseRfc1123Instant(lastModified)
         } catch (e: Exception) {
-            throw IllegalStateException("Invalid last-modified date format: $lastModified", e)
+            throw IllegalStateException("Invalid release date format: $lastModified", e)
         }
-        return releasedAt
+    }
+
+    protected fun parseRfc1123Instant(value: String): Instant {
```
nikitavlaevjb: The name is wrong now; Split into two functions and add them according to the usecase?

app/src/main/kotlin/io/klibs/app/service/impl/CentralSonatypeUserIndexingRequestService.kt
dkrasnoff: Why?
I don't wont indexing enabled by default on my local machine

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientRedirectTest.kt
```diff
@@ -321,20 +362,13 @@ class BaseMavenSearchClientRedirectTest {
         logger = LoggerFactory.getLogger(TestClient::class.java),
         objectMapper = ObjectMapper(),
         clientTransport = transport,
-        clock = clock
+        clock = clock,
+        lastModifiedHeader = "x-goog-meta-last-modified-epoch"
```
nikitavlaevjb: This is weird here, it will produce google epoch heder with RFC_1123_DATE_TIME, which never happens?
Central sonatype gives RFC_1123_DATE_TIME, google gives millis?

app/src/main/kotlin/io/klibs/app/service/impl/BaseUserIndexingRequestService.kt
```diff
@@ -50,30 +45,16 @@ internal class DefaultUserIndexingRequestService(
...
-                "No Kotlin Multiplatform artifacts found for $groupId${
-                    artifactId?.let { ":$it" }.orEmpty()
-                }"
-            )
-        }
+        val foundPackages = searchForPackages(groupId, artifactId)
 
-        val artifactsToSave = searchResult
-            .map { it.toMavenArtifact() }
+        val artifactsToSave = foundPackages
```
nikitavlaevjb: This has no KMP filtering anymore, why?

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import kotlin.test.assertNotNull
+import org.springframework.test.context.TestPropertySource
+
+@SpringBootTest
+@ContextConfiguration(
+    classes = [
+        MavenIntegrationConfiguration::class,
+        BaseMavenSearchClientTest.MavenStaticDataProviderTestConfiguration::class]
+)
+@TestPropertySource(
```
nikitavlaevjb: @ActiveProfiles("test")?

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientTest.kt
```diff
@@ -0,0 +1,58 @@
...
+        "klibs.integration.maven.central.rateLimitRefillAmount=100",
+        "klibs.integration.maven.central.rateLimitRefillPeriodSec=5",
+        "klibs.integration.maven.central.indexEndpoint=http://localhost:8080/index",
+        "klibs.integration.maven.central.indexDir=/tmp/maven-index",
+        "klibs.integration.maven.central.contentEndpoint=http://localhost:8080/content/",
+        "klibs.integration.maven.central.contentFallbackEndpoint=http://localhost:8080/fallback/",
+    ]
+)
+
+class BaseMavenSearchClientTest {
```
nikitavlaevjb: abstract?

app/src/test/resources/application-test.yml
```diff
@@ -1,6 +1,11 @@
 klibs:
   indexing: false
   ai: false
+  indexing-configuration:
+    central-sonatype:
+      enabled: true
```
nikitavlaevjb: Maybe enable only for some tests, no need to do it by dfeault?

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/GoogleMavenCentralMirrorSearchClient.kt
nikitavlaevjb: I meant something like this, factory methods
```
  @Configuration                                                      
  class MavenCentralSearchClientConfiguration {                       
                                                                      
      @Bean("CENTRAL_SONATYPE")                                       
      @ConditionalOnProperty("klibs.indexing-configuration.central-son
  atype.enabled", havingValue = "true")                               
      fun centralSonatypeSearchClient(                                
          xmlMapper: XmlMapper,                                       
          rateLimiter: MavenCentralRateLimiter,                       
          objectMapper: ObjectMapper,                                 
          props: MavenCentralProperties,                              
      ) = MavenCentralSearchClient(                                   
          xmlMapper, rateLimiter, objectMapper,                       
          contentEndpoint = props.contentEndpoint,                    
          contentFallbackEndpoint = props.contentFallbackEndpoint,    
          lastModifiedHeader = "last-modified",                       
      )                                                               
                                                                      
      @Bean("GOOGLE_MAVEN_CENTRAL_MIRROR")                            
      @ConditionalOnProperty("klibs.indexing-configuration.google-mave
  n-central-mirror.enabled", havingValue = "true")                    
      fun googleMavenCentralMirrorSearchClient(/* same args */) =     
  MavenCentralSearchClient(                                           
          /* ... */                                                   
          lastModifiedHeader = "x-goog-meta-last-modified-epoch",     
      )                                                               
  } 
  ```

**Review by nikitavlaevjb — APPROVED** [review]

File: app/src/main/kotlin/io/klibs/app/service/impl/CentralSonatypeUserIndexingRequestService.kt (hunk 1 above)
Comment (nikitavlaevjb) [reply app/src/main/kotlin/io/klibs/app/service/impl/CentralSonatypeUserIndexingRequestService.kt]: Otherwise it won't start?

File: integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/GoogleMavenCentralMirrorSearchClient.kt (hunk 3 above)
Comment (dkrasnoff) [reply integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/GoogleMavenCentralMirrorSearchClient.kt]: Got it, thanks for elaboration!
In such a case I don't see big difference between component

File: integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientRedirectTest.kt (hunk 18 above)
Comment (dkrasnoff) [reply integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientRedirectTest.kt]: Actually, google returns 3 types of `last modified` headers:

```
x-goog-meta-last-modified-epoch: 1788205020844
x-goog-meta-last-modified: Mon, 31 Aug 2026 19:37:00 GMT
last-modified: Tue, 01 Sep 2026 02:12:41 GMT
```
`last-modified` is, if I understand correctly, means modified in google repos (this one doesn't suit us).
`x-goog-meta-last-modified-epoch` and `x-goog-meta-last-modified` return the same date as real Maven Central `last-modified` header, but `x-goog-meta-last-modified` is not really a RFC 1123 format, because it can have date like `Mon, 24 Aug 2026  1:29:19 GMT`, that is why it is more realiable to use millis format

app/src/main/kotlin/io/klibs/app/service/impl/BaseUserIndexingRequestService.kt
dkrasnoff: Yes, we actually can skip it and it will happen in actual package indexing algorithm.
But here we can still check in banned or in queued packages.

nikitavlaevjb: But then some jvm package will be stuck in our indexing request table, why?
Also I would like it to fail fast and return 400 for the user here or smth

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientRedirectTest.kt
nikitavlaevjb: Yes, so x-goog-meta-last-modified-epoch with RFC_1123_DATE_TIME still never happens, right?

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/GoogleMavenCentralMirrorSearchClient.kt
nikitavlaevjb: The only small win for the methods here is that all similar things will be near each other in one class.
Otherwise yeah, its kinda the same, so feel free to ignore it

app/src/main/kotlin/io/klibs/app/service/impl/BaseUserIndexingRequestService.kt
dkrasnoff: > But then some jvm package will be stuck in our indexing request table, why?

It will be addressed in KTL-2087 . I'd say it is ok for us to collect such package as well.

> Also I would like it to fail fast and return 400 for the user here or smth

They will get the error, but not immediately and it is ok. It is batch request and we shouldn't immediately check all the version on each request. 
Also, I expect that most of the users will ping us only for real kmp packages

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientRedirectTest.kt
dkrasnoff: Yes, `x-goog-meta-last-modified-epoch` returns only millis, `x-goog-meta-last-modified` returns broken RFC 1123

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/GoogleMavenCentralMirrorSearchClient.kt
dkrasnoff: Thanks, exactly this change I'll ignore then

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientRedirectTest.kt
nikitavlaevjb: And test pairs x-goog-meta-last-modified-epoch with RFC time, which should not happen

app/src/test/resources/application-test.yml
dkrasnoff: removed

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientTest.kt
dkrasnoff: fixed

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientTest.kt
dkrasnoff: fixed

integrations/maven/src/test/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClientRedirectTest.kt
dkrasnoff: I've refactored code and test

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/impl/BaseMavenSearchClient.kt
dkrasnoff: splitted

integrations/maven/src/main/kotlin/io/klibs/integration/maven/search/MavenSearchClient.kt
dkrasnoff: removed

app/src/main/kotlin/io/klibs/app/service/impl/CentralSonatypeUserIndexingRequestService.kt
dkrasnoff: Implemented `klibs.indexing-configuration.central-sonatype.type` which allows to define non or exactly one type of the Maven Central discoverer at the same time.
Still for user requests Central Sonatype discoverer will always work.

PR #439, opened 2026-09-02
KTL-4909: create klibs-io-notifier gradle plugin

zwiora:


dkrasnoff (commented):
As it is a separate module, it needs a personal README.md file.
Which should at least includes next information blocks:
1. Short explanation about what this module
2. Instructions on how to apply this plugin to your project (with settings description if any custom settings are possible)

zwiora:
@dkrasnoff I've added the readme and also updated the FAQ section

notifier-plugin/README.md
```diff
@@ -0,0 +1,168 @@
...
+  - [apiBaseUrl](#apibaseurl)
+- [How it works](#how-it-works)
+- 
+## Prerequisites
+
+To use the klibs-io-notifier plugin in your project, it needs to:
+
+- be a Kotlin Multiplatform library
+- use Gradle (Kotlin Toolchain support coming soon) 
+- use the `com.vanniktech.maven.publish` plugin for publishing to Maven Central
```
dkrasnoff: Why should we stick to the `com.vanniktech.maven.publish` and not to official MavenPublishPlugin? 
As I understand we can just change `project.plugins.withId("com.vanniktech.maven.publish")` to `project.plugins.withType(MavenPublishPlugin::class.java)` and users will not be required to use `com.vanniktech.maven.publish`

dkrasnoff (changes requested):

notifier-plugin/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+```bash
+./gradlew publishAndReleaseToMavenCentral
+```
+> [!CAUTION]
+> Artifacts released to Maven Central cannot be deleted or modified ([read more](https://central.sonatype.org/faq/can-i-change-a-component/)). Before publishing your project in auto-release mode, make sure everything is fully tested and the artifacts are exactly as you intend them to be.
+
+### 5. (optional) Check the indexing status of your library
+
+After successfully notifying klibs.io, you can check the indexing status of your artifact at `https://klibs.io/package/<GROUP_ID>/<ARTIFACT_ID>/<VERSION>/status`.
```
dkrasnoff, line 118: It seems, that you need update frontend as well, as now it doesn't have such endpoint, only backend has

notifier-plugin/README.md
```diff
@@ -0,0 +1,168 @@
...
+```kotlin
+plugins {
+    id("io.klibs.notifier") version "1.0.0"
+}
+```
+
+### 4. Publish the library
+After a successful publishing task, the klibs-io-notifier automatically sends a notification. There is no separate command to run it. You just need to run one of the publishing tasks as usual.
+
+#### Manual release mode
```
dkrasnoff, line 86: I'd also make it possible to just call `notifyKlibsIo` like:
```
./gradlew notifyKlibsIo
```
and add it to instruction

notifier-plugin/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+After successfully notifying klibs.io, you can check the indexing status of your artifact at `https://klibs.io/package/<GROUP_ID>/<ARTIFACT_ID>/<VERSION>/status`.
+
+## Configuration
+
+The plugin exposes the following settings through the `klibsIoNotifier` extension.
+
+| Setting             | Description                                                                                     | Default |
+|---------------------|-------------------------------------------------------------------------------------------------| --- |
+| [apiBaseUrl](#apiBaseUrl) | Base URL of the klibs.io notification receiver. The plugin appends it with `/notify/artifacts`. | `https://klibs.io` |
```
dkrasnoff: I'd make it possible to fully replace the url, without adding `/notify/artifacts` postfix

notifier-plugin/README.md
zwiora: As I understand it, the official MavenPublishPlugin [does not support publishing to the new Maven Central Publishing Portal](https://docs.gradle.org/current/userguide/publishing_maven.html#maven_central), so users need to rely on other plugins. The `project.plugins.withType(MavenPublishPlugin::class.java)` gate would indeed work for other plugins too, but the problem is that the `publishTaskNames` list, which I use for selecting the task that klibs-io-notifier should hook into, is not universal. The tasks can be named differently in each plugin, so in these cases, the notifier would just never run

zwiora: I added an option for users to configure the task that KlibsIoNotifier will hook into via extensions. It will work for the Vanniktech plugin without additional configuration.

notifier-plugin/README.md
zwiora: done

notifier-plugin/README.md
zwiora: done

notifier-plugin/README.md
zwiora: done

frontend/next.config.mjs
```diff
@@ -6,6 +6,10 @@ const nextConfig = {
 				source: '/sitemap.xml',
 				destination: `${process.env.NEXT_PUBLIC_API_URL}/sitemap.xml`,
 			},
+			{
```
dkrasnoff, line 9: fyi @krutilov

notifier-plugin/src/test/kotlin/io/klibs/notifier/KlibsIoNotifierPluginFunctionalTest.kt
```diff
@@ -0,0 +1,374 @@
...
+        .withProjectDir(projectDir)
+        .withPluginClasspath()
+        .withArguments(argument)
+
+    @Test
+    @Timeout(60, unit = TimeUnit.SECONDS)
+    fun `notifies klibs with publication coordinates when publishing succeeds`() {
+        writeProject(disablePublishDependencies)
+
+        val result = runner("publishKotlinMultiplatformPublicationToMavenCentralRepository").build()
```
dkrasnoff: Create a constant for `publishKotlinMultiplatformPublicationToMavenCentralRepository`and use this constant everywhere in the code instead of copypasting.

notifier-plugin/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+<!-- TODO: KTL-4938 update information about Kotlin Toolchain-->
+
+### Minimum supported versions
+<!-- TODO: KTL-4909 Update the section according to the final decision -->
+
+* JDK 17
+* Gradle 9.0.0
+* Kotlin Gradle Plugin 2.2.0
+* vanniktech gradle-maven-publish-plugin 0.36.0
```
dkrasnoff: Is it still required?

dkrasnoff (approved):

notifier-plugin/README.md
zwiora: Not necessarily. I made it a suggestion, not a requirement now

notifier-plugin/src/test/kotlin/io/klibs/notifier/KlibsIoNotifierPluginFunctionalTest.kt
zwiora: done

dkrasnoff (approved):

PR #506, opened 2026-09-10
Updated new packages queue process job

dkrasnoff:


app/src/main/kotlin/io/klibs/app/configuration/properties/ProcessPackageIndexingQueueProperties.kt
```diff
@@ -0,0 +1,15 @@
+package io.klibs.app.configuration.properties
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+import java.time.Duration
+
+@ConfigurationProperties(ProcessPackageIndexingQueueProperties.PROCESS_PACKAGE_INDEXING_QUEUE_PREFIX)
+data class ProcessPackageIndexingQueueProperties(
+    val enabled: Boolean = true,
+    val deadline: Duration = Duration.ofMinutes(29),
+    val fixedRate: Duration = Duration.ofMinutes(30),
```
nikitavlaevjb, line 10: Maybe make it a single property and derive the second one?

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
nikitavlaevjb, line 159: Entity is detached here, logging will fail because `user_request_issue_id ` is lazy

nikitavlaevjb, line 143: This case will make the loop process the same row over and over.
IN_PROCESS prevented that

nikitavlaevjb (changes requested):

app/src/main/kotlin/io/klibs/app/configuration/properties/ProcessPackageIndexingQueueProperties.kt
dkrasnoff: Good catch, it was convenient to me during development to have this configuration flexible.
But for release branch I agree that we can leave only one property

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
dkrasnoff: Hm, didn't catch it during the testing, but I'll check, thanks!

dkrasnoff: REQUIRED_NEW

app/src/main/kotlin/io/klibs/app/configuration/properties/ProcessPackageIndexingQueueProperties.kt
dkrasnoff: Fixed (removed deadline)

app/src/main/kotlin/io/klibs/app/indexing/PackageIndexingService.kt
dkrasnoff: Fixed (introduced gav() method)

dkrasnoff: Fixed

nikitavlaevjb (approved):

PR #512, opened 2026-09-11
KTL-2087: Added error type for missing POM

dkrasnoff:
Also, removed `releaseTs` field from NonKmpPackageEntity, updated related services, exceptions, and tests.


core/package/src/main/kotlin/io/klibs/core/pckg/enums/PackageIndexingErrorType.kt
```diff
@@ -2,4 +2,5 @@ package io.klibs.core.pckg.enums
 
 enum class PackageIndexingErrorType {
     MISSING_TOOLING_METADATA,
+    MISSING_POM,
```
nikitavlaevjb, line 5: That raises the question on what is actually stored in the non_kmp_packages.
Are those even packages?
Artifacts?
Coordinates?

nikitavlaevjb (approved):

core/package/src/main/kotlin/io/klibs/core/pckg/enums/PackageIndexingErrorType.kt
dkrasnoff: Discussed in DM. 
Decided to stick to the next naming:
package -> package or kmp_package - directory on maven with VALID pom and json VALID
maven_artifact -> maven_coordinate - all coordinates from Maven Central
non_kmp_packages -> rejected_maven_coordinate - coordinates of rejected package + reason + time + scraper_type
package_index_request -> package_index_request - create a task for replacing it with queue


`non_kmp_packages` was renamed to `rejected_maven_coordinate` in the previous PR in the current stack.
`maven_artifact` is renamed to `maven_coordinate` in this PR

PR #521, opened 2026-09-14
KTL-2087: Added handling for minimized README with NUL bytes.

dkrasnoff:


core/project/src/main/kotlin/io/klibs/core/project/repository/ProjectRepositoryJdbc.kt
```diff
@@ -98,6 +103,10 @@ class ProjectRepositoryJdbc(
     }
 
     override fun updateMinimizedReadme(id: Int, minimizedReadme: String?) {
+        if (minimizedReadme?.contains('\u0000') == true) {
```
nikitavlaevjb, line 106: why can't we sanitize?
also, if we return here, we will still save the readme in s3 and has_readme = true, is that intended?

nikitavlaevjb (changes requested):

core/project/src/main/kotlin/io/klibs/core/project/repository/ProjectRepositoryJdbc.kt
dkrasnoff: It can, moved to the raw readme processing

core/readme/src/main/kotlin/io/klibs/core/readme/ReadmeContentBuilder.kt
```diff
@@ -22,24 +22,32 @@ class ReadmeContentBuilder(
...
-            raw = readmeMd,
+            raw = sanitizedReadme,
             markdown = processedMarkdownReadme,
             html = processedReadmeHtml,
             minimized = minimizedReadme,
         )
     }
 
+    private fun sanitizeReadme(readme: String): String {
+        if ('\u0000' in readme) return ""
```
nikitavlaevjb, line 48: Please leave comment why: that is wrong UTF and such

nikitavlaevjb (approved):

PR #523, opened 2026-09-14
KTL-2087: Added manual package indexing management API.

dkrasnoff:
Also, added processing service and related tests. Updated repositories to support batch operations.


app/src/main/kotlin/io/klibs/app/indexing/ManualPackageIndexProcessingService.kt
```diff
@@ -0,0 +1,282 @@
...
+            )
+        }
+    }
+
+    private fun deleteRequestsAndReportFailure(
+        packageIndexRequestsToDelete: List<IndexingRequestEntity>,
+        failureReason: String,
+    ) {
+        packageIndexRequestsToDelete.forEach { req ->
+            userRequestReportWriter.saveFailureReport(req.idNotNull, failureReason)
```
nikitavlaevjb, line 278: Why do we save it as a "user request failure"?
I understand that you wanted to have some history here, but it seems that if we delete all traces and leave only logs it would be enough

app/src/main/kotlin/io/klibs/app/indexing/ManualPackageIndexProcessingService.kt
```diff
@@ -0,0 +1,282 @@
...
+    private fun processMarkNonKmp(
+        pkg: PackageCoordinatesDTO,
+        errorType: PackageIndexingErrorType?,
+        reason: String?,
+    ): ProcessPackageIndexResultDTO {
+        val effectiveErrorType =
+            errorType ?: throw IllegalArgumentException("Error type must be specified for non-KMP package processing")
+
+        return if (pkg.version != null) {
+            processSingleNonKmpPackage(pkg, pkg.version, effectiveErrorType, reason)
```
nikitavlaevjb, line 121: whats the point?
We can move that `if` into `processMultipleNonKmpPackages` and delete `processSingleNonKmpPackage`?

app/src/main/kotlin/io/klibs/app/indexing/ManualPackageIndexProcessingService.kt
```diff
@@ -0,0 +1,282 @@
...
+                status = ProcessResultStatus.NOT_FOUND,
+                action = ProcessPackageIndexAction.REMOVE,
+                message = "No matching package index requests found",
+                affectedRequestsCount = 0,
+            )
+        }
+
+        deleteRequestsAndReportFailure(foundPackageIndexRequests, reason ?: "Removed manually by administrator")
+
+        logger.info(
```
nikitavlaevjb, line 96: I dont like that we log success for every package and on the last package we can get exception and everything will rollback

app/src/main/kotlin/io/klibs/app/indexing/ManualPackageIndexProcessingService.kt
```diff
@@ -0,0 +1,282 @@
...
+    private fun saveNonKmpPackagesAndCleanQueue(
+        packageCoordinatesDto: PackageCoordinatesDTO,
+        versions: List<String>,
+        foundPackageIndexRequests: List<IndexingRequestEntity>,
+        effectiveErrorType: PackageIndexingErrorType,
+        reason: String?,
+    ) {
+        versions.forEach { version ->
+            val scraperType =
+                foundPackageIndexRequests.firstOrNull { it.version == version }?.repo ?: ScraperType.CENTRAL_SONATYPE
```
nikitavlaevjb, line 235: Why fallback on sonatype?

nikitavlaevjb (approved):

app/src/main/kotlin/io/klibs/app/indexing/ManualPackageIndexProcessingService.kt
dkrasnoff: Logs live not to long, sometimes we could return much after the incident, we wanted to preserve it.

app/src/main/kotlin/io/klibs/app/indexing/ManualPackageIndexProcessingService.kt
dkrasnoff: Make sense to make it error I think, you are right

app/src/main/kotlin/io/klibs/app/indexing/ManualPackageIndexProcessingService.kt
dkrasnoff: And this is not only about the log, it is user request report, we notify user based on it as well

app/src/main/kotlin/io/klibs/app/indexing/ManualPackageIndexProcessingService.kt
dkrasnoff: made processing one by one

app/src/main/kotlin/io/klibs/app/indexing/ManualPackageIndexProcessingService.kt
dkrasnoff: Fixed

app/src/main/kotlin/io/klibs/app/indexing/ManualPackageIndexProcessingService.kt
dkrasnoff: Fixed

PR #533, opened 2026-09-16
ktl-4701 fix: fix breadcrumbs issue

krutilov:


frontend/src/app/ui/breadcrumb/index.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+        packageVersions={versions}
+        parentProject={projectDetails()}
+        version={version}
+    />,
+);
+
+describe('PackageBreadcrumbs', () => {
+    test('shows the latest released version when the url has none', () => {
+        renderBreadcrumbs([
+            packageOverview({ id: 1, version: '1.10.2', releasedAtMillis: 1_700_000_000_000 }),
```
nikpachoo: I can again offer the same: using fixture properties instead of hardcoding strings.
Not a blocker; just see that this may make the code cleaner.

nikpachoo (approved):

frontend/src/app/ui/breadcrumb/index.test.tsx
krutilov: moved package data to fixtures

PR #537, opened 2026-09-18
KTL-4787: Ban Suspicious Github Forks

Dan4Life:
Added the spec for the KTL-4787 ticket.


docs/specs/suspicious-fork-auto-ban/spec.md
```diff
@@ -0,0 +1,108 @@
...
+- **When:** the ban job runs.
+- **Then:** nothing is banned and the row keeps the reviewer's `status` and `notes`.
+- **Independent test:** resolve a qualifying candidate by hand, run the job, assert no ban row and the row unchanged.
+
+### Edge cases
+- **The fork was renamed or deleted.** Neither is found, so the candidate stays `PENDING` — a miss costs a manual review, not a wrong ban.
+- **GitHub is unavailable or rate-limited.** No decision is recorded; a failed lookup is never read as "not a fork".
+- **The candidate no longer conflicts.** Candidate rows are kept after a conflict disappears, so a row can describe a state that no longer holds. Banning one entry also deletes its `package` rows, which can leave siblings under a single `groupId` and drop them out of detection mid-run. The conflict set is therefore taken once at the start of a run: otherwise, where every member of a conflict is a suspect — 2 cases in the production copy, both of them one publisher using a flat and a nested namespace for the same library — only whichever happened to be processed first would be banned.
```
nikitavlaevjb: Ive read this 4 times and cannot understand a thing.
Looks like claudish language, too complicated. Rewrite and simplify, probably needs some structure instead of complex sentences.

nikitavlaevjb (approved):
Almost LGTM!

app/src/main/kotlin/io/klibs/app/service/ForkChecker.kt
```diff
@@ -0,0 +1,35 @@
+package io.klibs.app.service
+
+import io.klibs.integration.github.GitHubIntegration
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+
+@Service
+class ForkChecker(
```
nikitavlaevjb: Why is that a separate service?

core/package/src/main/kotlin/io/klibs/core/pckg/repository/SuspiciousPackageCandidateRepository.kt
```diff
@@ -29,4 +31,66 @@ interface SuspiciousPackageCandidateRepository :
...
+            FROM suspicious_package_candidate candidate
+                     JOIN (SELECT project_id, artifact_id, group_id,
+                                  count(*) OVER (PARTITION BY project_id, artifact_id) AS group_id_count
+                           FROM (SELECT DISTINCT project_id, artifact_id, group_id
+                                 FROM package
+                                 WHERE project_id IS NOT NULL) distinct_entries) entry
+                          ON entry.project_id = candidate.project_id
+                              AND entry.artifact_id = candidate.artifact_id
+                              AND entry.group_id = candidate.group_id
+                              AND entry.group_id_count > 1
```
nikitavlaevjb: Why need entry.group_id_count > 1?

integrations/github/src/main/kotlin/io/klibs/integration/github/GitHubIntegrationKohsukeLibrary.kt
```diff
@@ -377,6 +377,8 @@ internal class GitHubIntegrationKohsukeLibrary(
             stars = this.stargazersCount,
             openIssues = this.openIssueCount,
             lastActivity = this.pushedAt.toInstant(),
+            fork = this.isFork,
+            parentFullName = this.parent?.fullName,
```
nikitavlaevjb: `parent?.fullName` is not a plain field read. In github-api 1.330:

```java
public GHRepository getParent() throws IOException {
    if (fork && parent == null) {
        populate();   // unconditional GET /repos/{full_name}
    }
    ...
}
```

`populate()` has no already-populated guard, so for any fork whose JSON did not carry `parent` this does a synchronous extra API call — and `toModel()` is invoked *outside* `executeNullable` / `executeNullableWithAnonymousFallback` (lines 75 and 85). That means the hidden request is not timed into `klibs.github.request.time`, not mapped to `null` on 404, and not covered by the anonymous IP-allow-list retry.

Today both callers fetch the full `/repos/...` payload, which does include `parent`, so it should not fire in practice. But it is one narrow `GHRepository` away from firing on the main indexing path. Please move `.toModel()` inside the guarded block so every field access on `GHRepository` is covered.

app/src/main/kotlin/io/klibs/app/job/BanSuspiciousForksJob.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.stereotype.Component
+import java.util.concurrent.TimeUnit
+
+@Component
+@ConditionalOnProperty("klibs.indexing", havingValue = "true")
+class BanSuspiciousForksJob(
+    private val suspiciousForkBanService: SuspiciousForkBanService,
+) {
+
+    @Scheduled(initialDelay = 0, fixedRate = 1, timeUnit = TimeUnit.DAYS)
```
nikitavlaevjb: `initialDelay = 0, fixedRate = 1 DAY` means this runs on every pod start, and since there is no `lockAtLeastFor` the ShedLock row is released as soon as the run finishes — so N restarts in a day give N runs. For a job that deletes `package` rows I would rather have one predictable daily run: `IndexNewPackagesJob` uses `@Scheduled(cron = "0 0 2 * * *")` for exactly this reason.

Related: `CollectSuspiciousPackageCandidatesJob` has the identical schedule, so on a cold start there is no ordering between collecting candidates and banning them. Not harmful (the ban job just sees fewer rows), but a cron for each with a gap makes the pipeline readable.

app/src/main/kotlin/io/klibs/app/service/SuspiciousForkBanner.kt
```diff
@@ -0,0 +1,33 @@
...
+            projectId = candidate.projectId,
+            artifactId = candidate.artifactId,
+            groupId = candidate.groupId,
+            notes = reason,
+        )
+        if (resolved == 0) {
+            return false
+        }
+
+        check(blacklistService.banPackage(candidate.groupId, candidate.artifactId, reason)) {
```
nikitavlaevjb: `check()` raises `IllegalStateException`, which means "this cannot happen". But `BlacklistService.banPackage` returns `false` for two perfectly normal situations — the coordinate is not in `package`, or it is already banned. So this uses an assertion as control flow.

The consequence is in the caller: `SuspiciousForkBanService.ban` catches `Exception` and buckets everything as `BAN_REFUSED`. A legitimate refusal and a real DB failure end up in the same counter, which defeats the "skipped-by-reason" requirement in spec §5 — a run where the DB is unhappy reports the same number as a run where nothing needed banning.

I can see why the order is resolve-then-ban (it avoids a ban with an unresolved candidate if a reviewer races you), so keep the order but roll back explicitly instead of throwing:

```kotlin
if (!blacklistService.banPackage(candidate.groupId, candidate.artifactId, reason)) {
    TransactionAspectSupport.currentTransactionStatus().setRollbackOnly()
    return false
}
```

Then the caller's `catch` only sees genuine failures and can count them separately.

app/src/main/kotlin/io/klibs/app/service/SuspiciousForkBanner.kt
```diff
@@ -0,0 +1,33 @@
+package io.klibs.app.service
+
+import io.klibs.core.pckg.dto.projection.ForkBanCandidateView
+import io.klibs.core.pckg.repository.SuspiciousPackageCandidateRepository
+import io.klibs.core.project.blacklist.BlacklistService
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+
+@Service
+class SuspiciousForkBanner(
```
nikitavlaevjb: Naming: our convention (AGENTS.md) is `*Service` / `*Repository` / `*Controller`. `SuspiciousForkBanner` is a new suffix.

More generally, this feature is ~145 lines of production code spread over three classes — `ForkChecker`, `SuspiciousForkBanService`, `SuspiciousForkBanner`. Only one of them *has* to be separate: this one, because `@Transactional` would not proxy on a self-call. `ForkChecker` (see my other thread) has no such reason.

Suggestion: fold `ForkChecker` into `SuspiciousForkBanService`, and either rename this to something like `ForkBanTransactionService` or, better, move the transactional unit down into `core/package` / `core/project` where the two repositories it touches already live.

docs/specs/suspicious-fork-auto-ban/spec.md
```diff
@@ -0,0 +1,111 @@
...
+- **Rejected:** `project.dependent_count`, which cannot express a coordinate-level fact at any granularity; and storing a per-candidate count, which would need keeping in sync.
+- **Limitation:** `package_dependency` only holds dependencies declared by packages klibs has indexed — 290,278 of 551,881 packages contribute any edges. An application or unindexed consumer is invisible, so this establishes *no known dependants*, never *no dependants*. It is one-sided: a hit blocks the ban, a miss is not proof of safety. A cheap veto, not a guarantee — the fork check is what authorises the ban.
+
+### Decision — Fork confirmation is a direct repository lookup
+- **Choice:** look up `<suspect owner>/<repository name>` and require it to be a fork whose parent is the repository the candidate resolves to.
+- **Why:** one call per pair, reusing `getRepository(owner, name)` with two new fields. A fork normally keeps the original's name.
+- **Rejected:** enumerating the owner's repositories and matching on parent — survives renames, but costs several paginated calls per owner.
+- **Revisit if:** renamed forks prove common; a miss currently costs only a manual review.
+
+## 11. Test strategy
```
nikitavlaevjb: Section numbering jumps 8 → 11. Either 9 and 10 were dropped from the template and the rest should be renumbered, or they are genuinely missing.

docs/specs/suspicious-fork-auto-ban/spec.md
```diff
@@ -0,0 +1,111 @@
...
+- **Given:** a candidate a reviewer has already set to `RESOLVED`, which nonetheless meets every condition for an automatic ban.
+- **When:** the ban job runs.
+- **Then:** nothing is banned and the row keeps the reviewer's `status` and `notes`.
+- **Independent test:** resolve a qualifying candidate by hand, run the job, assert no ban row and the row unchanged.
+
+### Edge cases
+- **The fork was renamed or deleted.** Neither is found, so the candidate stays `PENDING` — a miss costs a manual review, not a wrong ban.
+- **GitHub is unavailable or rate-limited.** No decision is recorded; a failed lookup is never read as "not a fork".
+- **The candidate no longer conflicts.** Candidate rows stay in the database after their conflict disappears. Each run therefore checks which candidates still conflict — the same `artifactId` under more than
+  one `groupId` in one project — and skips the rest, leaving `status` and `notes` unchanged.
```
nikitavlaevjb: Worth spelling out the operational consequence of this rule. Once one member of a conflict is banned, its sibling's candidate row keeps `status = PENDING` but can never qualify again (the `group_id_count > 1` gate excludes it) and nothing will ever resolve it. So the manual review queue accumulates rows that are permanently undecidable by both the job and, arguably, a human.

That may be acceptable, but the spec should say so — and if it is not acceptable, the fix belongs here rather than being discovered later when the queue stops shrinking.

core/package/src/main/kotlin/io/klibs/core/pckg/repository/SuspiciousPackageCandidateRepository.kt
nikitavlaevjb: Is it because of ban? If we ban, and conflict is resolved, maybe we should remove the "not longer conflicting" record?

app/src/main/kotlin/io/klibs/app/service/SuspiciousForkBanner.kt
nikitavlaevjb: Sorry for slop, but this is mostly correct.

nikitavlaevjb (changes requested):

